Issue #1, opened 2018-01-31
Provide a setup.py file


sahuguet:
By packaging the api-client the correct, we can install our SDK using `pip` from github.

See https://github.com/pypa/pip/issues/3610,
https://docs.python.org/2/distutils/setupscript.html

PR #29, opened 2019-01-10
Add setup.py for pip install

itsjohnward:
#1 

Can test this using

```sh
pip install git+https://github.com/gro-intelligence/api-client.git@CLEWS-14821
```

I have been creating & deleting virtualenvs to test like

```sh
virtualenv req
. req/bin/activate
pip install git+https://github.com/gro-intelligence/api-client.git@CLEWS-14821
...
deactivate
rm -rf req
```

Note you can now use the `gro` command from anywhere on the command line instead of navigating to api/client and using `python gro_client.py`. See README instructions


itsjohnward:
Note: one thing this does not do is expose the CropModel class if installed via pip. Need to move it out of the samples folder.

nemozen (approved):

PR #30, opened 2019-02-12
more specific search query  for yield metrics

nemozen:
in this example we want to guarantee the top result is the expected metric

### DIFF
--- a/api/client/samples/crop_models/soybeans.py
+++ b/api/client/samples/crop_models/soybeans.py
@@ -26,7 +26,7 @@ def add_brazil_soybeans_yield(model):
     entities = {}
     # Search for item and metric by name, and use the top result
     entities['item_id'] =  model.search_for_entity('items', "soybeans")
-    entities['metric_id'] = model.search_for_entity('metrics', "yield")
+    entities['metric_id'] = model.search_for_entity('metrics', "yield mass/area")
     # There are many regions with "brazil" in the name, level = 3 is
     # the country
     for region in model.search_and_lookup('regions', "brazil"):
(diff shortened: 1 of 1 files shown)


JGrinst (approved):
LGTM

itsjohnward (changes requested):
This search matching is working as intended now, but I'm getting a divide-by-zero error in crop_model's compute_weights function. This line: `[old_div(mean,total) for mean in means]`

Looking into that now

itsjohnward (approved):
It was matching on the "Production Quantity" group instead of "production quantity (mass)"

I removed parens from

```
model.compute_crop_weighted_series(
        'soybeans', 'Production Quantity (mass)',
        'Vegetation (NDVI)', 'Vegetation Indices (index)',
        provinces)
```

Changed to:

```
model.compute_crop_weighted_series(
        'soybeans', 'Production Quantity mass',
        'Vegetation NDVI', 'Vegetation Indices index',
        provinces)
```

and that works. Should just working after Simon's special characters search change though, so no change should be necessary here on the client side.

nemozen:
> It was matching on the "Production Quantity" group instead of "production quantity (mass)"
> 
> I removed parens from
> 
> ```
> model.compute_crop_weighted_series(
>         'soybeans', 'Production Quantity (mass)',
>         'Vegetation (NDVI)', 'Vegetation Indices (index)',
>         provinces)
> ```
> Changed to:
> 
> ```
> model.compute_crop_weighted_series(
>         'soybeans', 'Production Quantity mass',
>         'Vegetation NDVI', 'Vegetation Indices index',
>         provinces)
> ```
> and that works. Should just working after Simon's special characters search change though, so no change should be necessary here on the client side.


Thanks! Yes search input escaping should fix it but might as well add the workaround there anyway, it doesn't hurt.  Done.

PR #32, opened 2019-02-14
Update setup.py to find and include all packages (including samples/crop_models) and update metric name

itsjohnward:
sugar.py's crop_model import doesn't work with pip install instructions. Have to add crop_model to the path. Fixed by including all packages automatically

rename CLI from `gro` to `gro_client`

"Production Quantity" matching on the exact name group "Production Quantity" instead of the metric "Production Quantity (mass)." Updated the command to reflect the stronger match.


JGrinst (approved):
Makes sense to me!

nemozen:
changed to make it install *all* packages including the samples, so we no longer need to remove teh example frmo teh README. The example should work now

nemozen (approved):
retested with fresh user

`pip install --target=/Users/tester git+https://github.com/gro-intelligence/api-client.git@update-readme`

PR #37, opened 2019-02-25
Discoverability

nemozen:
Improve client for discoverability

- Move general ontology and data series discovery features from sample module  up to main client module,
- Make separate crop model module for crop weighting functions

The changes are backward compatible with existing client code.


api/client/gro_client.py
```diff
@@ -21,62 +29,134 @@
...
+        the given selection of entities.
+        """
+        data_series_list = self.get_data_series(**selected_entities)
+        if not data_series_list:
+            raise Exception("No data series available for {}".format(
+                selected_entities))
+        selected_data_series = data_series_list[int(len(data_series_list)*random())]
+        return selected_data_series
+
+    def print_one_data_series(self, data_series, filename):
```
eyvindn, line 156: called "print" but outputs to file.  call this save_one_data_series or something similar?

api/client/gro_client.py
```diff
@@ -21,62 +29,134 @@
...
+        self._data_series_list.append(data_series)
+        self._data_series_queue.append(data_series)
+        self._logger.info("Added {}".format(data_series))
+        return
+
+    def add_data_series(self, **kwargs):
+        """Search for entities matching the given names, find data series for
+        the given combination, and add them to this objects list of
+        series."""
+        MAX_RESULTS=3
```
eyvindn: let's have max_results as a parameter

api/client/crop_model.py
```diff
@@ -0,0 +1,71 @@
...
+        for region in regions:
+            entities['region_id'] = region['id']
+            for data_series in self.get_data_series(**entities):
+                self.add_single_data_series(data_series)
+                break
+        # Compute the average over time for reach region
+        df = self.get_df()
+        
+        def mapper(region):
+            return df[(df['metric_id'] == entities['metric_id']) & \
```
eyvindn: should this be indexed by item_id as well, if we do weighting by a different crop?

api/client/crop_model.py
```diff
@@ -0,0 +1,71 @@
...
+                      (df['region_id'] == region['id'])]['value'].mean(skipna=True)
+        means = list(map(mapper, regions))
+        self._logger.debug('Means = {}'.format(
+            list(zip([region['name'] for region in regions], means))))
+        # Normalize into weights
+        total = math.fsum([x for x in means if not math.isnan(x)])
+        return [old_div(mean,total) for mean in means]
```
eyvindn: do we need old_div here or is it an artefact of making the code py3 compatible. `total` is a float so both py2 and py3 should do float division.

api/client/gro_client.py
```diff
@@ -21,62 +29,134 @@
...
-        print("Randomly selected item: {}".format(item['name']))
-        selected_entities = {'itemId':  item['id']}
-        entity_list = client.list_available(selected_entities)
-        num = len(entity_list)
-    entities = entity_list[int(num*random())]
-    print("Using entities: {}".format(str(entities)))
-    selected_entities.update(entities)
-    return selected_entities
+
+    _logger = api.client.lib.get_default_logger()
```
eyvindn: Are these intentionally static class variables (outside of an __init__)? Could cause confusion if there's ever two instances of GroClient() and one happens to call get_df before the other.

api/client/gro_client.py
eyvindn: or actually maybe not as this explodes if it's too big. but also an arbitrary number we've picked for our users.

nemozen: yeah we don't want users to change it unless they really really want to. moved to cfg

api/client/crop_model.py
nemozen: done

api/client/gro_client.py
nemozen: added a TODO to rename, since that would be non-backward compatible

api/client/gro_client.py

api/client/crop_model.py
nemozen: hmm yeah  this was [added for py3 compatibility](https://github.com/gro-intelligence/api-client/pull/23/files#diff-e2b5461a0cffb216bfd21e0c95b60222R133)  but not actually  necessary.
I've removed old_div. mean is already a float but I made that  explicit.

api/client/gro_client.py
```diff
@@ -1,82 +1,167 @@
...
+        super(GroClient, self).__init__(api_host, access_token)
+        self._logger = lib.get_default_logger()
+        self._data_series_list = []  # all that have been added
+        self._data_series_queue = []  # added but not loaded in data frame
+        self._data_frame = None
+
+    ###
+    # Finding, indexing and loading multiple data series into a data frame
+    ###
+    def get_df(self):
```
itsjohnward, line 54: Wrapping these functions in a class is a breaking change.

The [cash_crops.py](https://github.com/gro-intelligence/api-client/blob/4b6116c51d41e1b20ebbeecad4d8b30d69d9ccb6/api/client/samples/crop_models/cash_crops.py#L25) example demonstrated importing functions from gro_client.

So this is failing now with

```
  File "cash_crops.py", line 25, in <module>
    from api.client.gro_client import get_df
ImportError: cannot import name get_df
```

itsjohnward (changes requested):

api/client/gro_client.py
nemozen: done, re-added it as a deprecated method.

itsjohnward: Verified this fixes the cash_crops sample. Someone may have followed our example and imported other functions from there though. Need to do this for all of them?

`search_for_entity()`, `print_random_data_series()`, and `pick_random_entities()`

nemozen: done

itsjohnward (approved):

PR #39, opened 2019-03-11
Change import from package

nemozen:


itsjohnward (approved):
Relative import worked fine on test set with Python 2.7.15 but broke on Python 3. Confirmed using `from package` works on 3.7.0

eyvindn (approved):

PR #42, opened 2019-03-19
Changing sample series

JGrinst:
The sample series we had was rather ridiculous. Changed it to something more relevant.

### DIFF
--- a/api/client/samples/quick_start.py
+++ b/api/client/samples/quick_start.py
@@ -9,9 +9,9 @@ ACCESS_TOKEN=os.environ['GROAPI_TOKEN']
 def main():
     client = api.client.Client(API_HOST, ACCESS_TOKEN)
 
-    selected_entities = { u'region_id': 1038, # Cape Verde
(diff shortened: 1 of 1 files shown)


api/client/samples/quick_start.py
```diff
@@ -9,9 +9,9 @@
 def main():
     client = api.client.Client(API_HOST, ACCESS_TOKEN)
 
-    selected_entities = { u'region_id': 1038, # Cape Verde
-                          u'item_id': 5187, # Management of donkey manure
-                          u'metric_id': 5590032 } # Total Emissions Quantity (mass)
+    selected_entities = { u'region_id': 1029, # Brazil
```
nemozen: We already have brazil soybeans and brazil sugar examples in the crop models so this might get confusing...  Let's use something else here so it's easy to refer to them

nemozen (changes requested):

eyvindn:
lgtm

nemozen (approved):

PR #43, opened 2019-03-26
Add r usage example

itsjohnward:
Add proof of concept example for using the Python client from within R.

Important things to note:

- The Python client is the officially-supported method. This is a use-at-your-own-risk situation. See README for if that's clear or not.
- Python still needs to be installed in order to use this, since it calls Python from within R. It's not a replacement runtime - it's an interface wrapper.
- This is implemented as a Vagrantfile and provisioning script just to show what the pre-requisites and environment requirements are. This is one possible configuration using Ubuntu 16.04 and Python 2.7.12, but there are many other possible ways to fulfill the requirements. It is not an exhaustive instruction manual.

Click “view file” to see the formatted version of the README.

### DIFF
--- a/r-client/README.md
+++ b/r-client/README.md
@@ -6,22 +6,22 @@ What this Vagrantfile does is it creates an Ubunutu 16.04 virtual machine, insta
 
 This is implemented as a Vagrantfile to demonstrate exactly what dependencies are required in this specific test environment. This example is in no way saying you must or should use Vagrant if intending to use the Gro API in R yourself. It is also important to note that because R is not an officially supported language for using the Gro API, it may take extra configuration on the user's part, and not all features are tested.
 
-##Pre-requisites for running this example
+## Pre-requisites for running this example
 
 * vagrant
 * virtualbox
 
-##Preparation
+## Preparation
 
 In [src/testscript.R](src/testscript.R) replace the token string with your own token. See the README in the main repo here https://github.com/gro-intelligence/api-client for details on how to retrieve a token.
 
-##Run
+## Run
 
 ```sh
 $ vagrant up
 ```
 
-##To use this VM for testing your own R scripts
+## To use this VM for testing your own R scripts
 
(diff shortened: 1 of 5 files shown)


nemozen (approved):

eyvindn (approved):
LGTM. Tested locally and runs without issue. Still not convinced we should ship this out with a vagrant setup (feels like using a sledgehammer for a push-pin) but I suppose if we want it to work with very high reliability I can see the reasoning.

r-client/README.md
```diff
@@ -0,0 +1,46 @@
...
+This is implemented as a Vagrantfile to demonstrate exactly what dependencies are required in this specific test environment. This example is in no way saying you must or should use Vagrant if intending to use the Gro API in R yourself. It is also important to note that because R is not an officially supported language for using the Gro API, it may take extra configuration on the user's part, and not all features are tested.
+
+## Pre-requisites for running this example
+
+* vagrant
+* virtualbox
+
+## Preparation
+
+In [src/testscript.R](src/testscript.R) replace the token string with your own token. See the README in the main repo here https://github.com/gro-intelligence/api-client for details on how to retrieve a token.
```
eyvindn: We can link the readme directly (maybe even directly to that section) as we do for the testscript.R

itsjohnward: Done

itsjohnward:
> LGTM. Tested locally and runs without issue. Still not convinced we should ship this out with a vagrant setup (feels like using a sledgehammer for a push-pin) but I suppose if we want it to work with very high reliability I can see the reasoning.

Yeah, I agree. Wouldn't normally, but I ran into a couple problems with Python versions and went through a few iterations. So I created one for myself and then figured maybe someone else could benefit from it - take the working example and compare to their own setup. Specifically because it's intended as more of a proof-of-concept than a polished product, I don't think it's bad form to include it for reproducibility's sake.

One thing I did just do is re-shuffle the README so actually using it in R is split out and comes up above trying to run the vagrant example. More important thing first. See if that's any better.

PR #46, opened 2019-03-26
CLEWS-16132 Add automatic retries

itsjohnward:
Tested out the BatchClient and lib.py implementations by running test scripts on dev to generate as many requests as possible in many parallel shell processes:

For batch_client, I adapted `batch_client.py`:

```python
api_client = BatchClient(API_HOST, ACCESS_TOKEN)

    # specify everything except region_id
    selection = {
        'metric_id': 860032,
        'item_id': 274,
        'source_id': 25,
        'frequency_id': 9,
        'start_date': '1998-01-01T00:00:00.000Z',
        'end_date': '1998-01-01T00:00:00.000Z'
    }

    selections = []
    for region_id in US_COUNTY_IDS:
        selection_temp = dict(selection)
        selection_temp["region_id"] = region_id
        selections.append(selection_temp)

    # make the request in a batch, asynchronous way
    output = api_client.batch_async_get_data_points(selections)

    # the output data is in the same order as the input queries.
    for county_id, data in zip(US_COUNTY_IDS, output):
        print(data)
        # some counties are missing data
        if len(data) > 0 and 'data' in data[0]:
            print("county_idx=%i produced %.0f tonnes of corn in 1998" %
                  (county_id, data[0]["value"]))
        else:
            print("county_idx=%i has no data for 1998" % county_id)
```

Screen Shot 2019-03-26 at 8 57 03 PM

You can see 6 concurrent batch clients requesting data simultaneously. Without the retries, any instance of being rate limited would cause the program to fail, but with the rate limits, it waits a tick and tries again, and all 6 are able to resolve successfully, albeit with some slight delays.

And to test lib.py I used:

```python
    client = api.client.Client(API_HOST, ACCESS_TOKEN)
    i = 1
    start_time = time.time()
    num_requests = 0
    while True:
        num_requests += 1
        try:
            client.lookup('regions', i)
        except Exception as e:
            if str(e) == '429':
                print("Gave up")
            i += 1
        else:
            i += 1
        print(num_requests / (time.time() - start_time), "QPS")
```

Screen Shot 2019-03-26 at 10 00 20 PM

As you can see, running 11 processes concurrently, each one performing ~2.5-3qps, I was starting to run into some rate limiting, and some processes had to wait a tick or two before proceeding but continued with an occasional unlucky "Gave Up."

### DIFF
--- a/api/client/batch_client.py
+++ b/api/client/batch_client.py
@@ -49,6 +49,13 @@ class BatchClient(Client):
                                        connect_timeout=cfg.TIMEOUT)
             try:
                 data = yield self._http_client.fetch(http_request)
+                elapsed_time = time.time() - start_time
+                log_record = dict(base_log_record)
+                log_record['elapsed_time_in_ms'] = 1000 * elapsed_time
+                log_record['retry_count'] = retry_count
+                log_record['status_code'] = data.code
+                self._logger.debug('OK', extra=log_record)
+                raise gen.Return(data.body)
             except HTTPError as e:
                 elapsed_time = time.time() - start_time
                 log_record = dict(base_log_record)
@@ -67,14 +74,6 @@ class BatchClient(Client):
                     raise Exception('Giving up on {} after {} tries. \
                         Error is: {}.'.format(
                         url, retry_count, e.response.error))
-            else:
-                elapsed_time = time.time() - start_time
-                log_record = dict(base_log_record)
-                log_record['elapsed_time_in_ms'] = 1000 * elapsed_time
-                log_record['retry_count'] = retry_count
-                log_record['status_code'] = data.code
-                self._logger.debug('OK', extra=log_record)
-                raise gen.Return(data.body)
 
     @gen.coroutine
     def get_data_points(self, **selection):
--- a/api/client/batch_client.py
+++ b/api/client/batch_client.py
@@ -13,11 +13,7 @@ from tornado.escape import json_decode
 from tornado.httpclient import AsyncHTTPClient, HTTPRequest, HTTPError
 from tornado.ioloop import IOLoop
 from tornado.queues import Queue
-from tornado.locks import Event
 from api.client import cfg, lib, Client
-
-rate_limit_lock = Event()
-rate_limit_lock.set()
 
 
 class BatchClient(Client):
@@ -66,9 +62,7 @@ class BatchClient(Client):
                     self._logger.warning(e.response.error, extra=log_record)
                     retry_count += 1
                     if e.code == 429:
-                        rate_limit_lock.clear()
                         time.sleep(2 ** retry_count)  # Exponential backoff
-                        rate_limit_lock.set()
                 else:
                     self._logger.error(e.response.error, extra=log_record)
                     raise Exception('Giving up on {} after {} tries. \
--- a/api/client/batch_client.py
+++ b/api/client/batch_client.py
@@ -10,7 +10,7 @@ except ImportError:
 
 from tornado import gen
 from tornado.escape import json_decode
-from tornado.httpclient import AsyncHTTPClient, HTTPRequest
+from tornado.httpclient import AsyncHTTPClient, HTTPRequest, HTTPError
 from tornado.ioloop import IOLoop
 from tornado.queues import Queue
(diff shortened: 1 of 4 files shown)


api/client/batch_client.py
```diff
@@ -10,28 +10,33 @@
...
     def __init__(self, api_host, access_token):
         super(BatchClient, self).__init__(api_host, access_token)
         self._logger = lib.get_default_logger()
         self._http_client = AsyncHTTPClient()
 
     @gen.coroutine
     def get_data(self, url, headers, params=None):
-        """General 'make api request' function. Assigns headers and builds in retries and logging."""
+        """General 'make api request' function.
+
```
nemozen, line 33: extra line

api/client/batch_client.py
```diff
@@ -40,51 +45,68 @@ def get_data(self, url, headers, params=None):
...
+                    if e.code == 429:
+                        rate_limit_lock.clear()
+                        time.sleep(2 ** retry_count)  # Exponential backoff
+                        rate_limit_lock.set()
+                else:
+                    self._logger.error(e.response.error, extra=log_record)
+                    raise Exception('Giving up on {} after {} tries. \
+                        Error is: {}.'.format(
+                        url, retry_count, e.response.error))
+            else:
```
nemozen: why else? the following block should just be in the try block above

nemozen (changes requested):

api/client/batch_client.py
itsjohnward: Extra line is from PEP 257 https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

Was too long to be a one-line comment, so I split it up into summary and description lines

api/client/batch_client.py
itsjohnward: Done

nemozen (approved):

api/client/batch_client.py
```diff
@@ -40,51 +45,67 @@ def get_data(self, url, headers, params=None):
...
+                elapsed_time = time.time() - start_time
+                log_record = dict(base_log_record)
+                log_record['elapsed_time_in_ms'] = 1000 * elapsed_time
+                log_record['retry_count'] = retry_count
+                log_record['status_code'] = e.code
+                if retry_count < cfg.MAX_RETRIES:
+                    self._logger.warning(e.response.error, extra=log_record)
+                    retry_count += 1
+                    if e.code == 429:
+                        rate_limit_lock.clear()
```
eyvindn: the rate_limit_lock event is not being yielded (await in py3), so it's not doing anything. 

tornado is a single-thread async library so a vanilla python time.sleep() will actually block all execution. this is a fine assumption if we are just doing a batch async set of calls (we want a 429 event to stop all our requests that are in flight). **i propose remove the lock for now and just keep the sleep part here**

however if someone wants to combine our async calls with other async coroutines then we should indeed 
```_ = yield rate_limit_lock```
here I believe, and make sure it's initially set, then follow it with a [tornado.gen.sleep](https://www.tornadoweb.org/en/stable/gen.html#tornado.gen.sleep)
**but let's do this down the line**

an additional issue is that if we have multiple requests in flight receiving a 429 we want all of them to wait, instead of one by one processing this backoff, i.e. maybe having a shared variable across the co-routines. but this is even less of a priority.

eyvindn (changes requested):
i propose remove the lock for now as it's not doing anything. otherwise LGTM

api/client/batch_client.py
itsjohnward: Ah I didn't realize sleep would block all coroutines. Will remove.

Regarding your point about the multiple requests in flight, I think this is what you were getting at, but to try to clarify: several coroutines could potentially run into a rate limit within the same period before the first response is received, so ideally we could consider request start time and consolidate 429s within the same period and just do one backoff instead of several in a row when processing a backlog of responses. Small optimization that only really applies to multiple requests within the amount of time it takes to respond a 429 - shouldn't be prevalent unless there's significant latency.

eyvindn: correct. yep let's leave that aside as premature optimization for now.

eyvindn (approved):

PR #47, opened 2019-03-28
Lint and add docstrings to lib module

itsjohnward:
This PR does three things:

1. Modify code style so the lib.py module is compliant with flake8, pep8, pylint, and pydocstyle
2. Add numpy-style docstrings to all functions in the lib.py module https://numpydoc.readthedocs.io/en/latest/format.html#documenting-constants
3. Add doctest unit tests to all functions that do not need any API mocking

Run tests using `python lib.py -v` as noted in the comment here: https://github.com/gro-intelligence/api-client/compare/lint-lib?expand=1#diff-7ed38827a2c4e58f0dd776f8639dc95cR820

### DIFF
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -11,6 +11,15 @@ CROP_CALENDAR_METRIC_ID = 2260063
 
 
 def get_default_logger():
+    """Get a logging object using the default log level set in cfg.
+
+    https://docs.python.org/3/library/logging.html
+
+    Returns
+    -------
+    logger : logging.Logger
+
+    """
     logger = logging.getLogger(__name__)
     logger.setLevel(cfg.DEFAULT_LOG_LEVEL)
     if not logger.handlers:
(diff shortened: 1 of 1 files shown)


api/client/lib.py
```diff
@@ -1,46 +1,95 @@
...
 
+REGION_LEVELS = {
+    'world': 1,
+    'continent': 2,
+    'country': 3,
+    'province': 4,  # Equivalent to state in the United States
+    'district': 5,  # Equivalent to county in the United States
+    'city': 6,
+    'market': 7,
+    'group': 8,  # arbitrary grouping of regions of any level
```
itsjohnward: @ reviewers double-check this. Want to use just "group"? Or "regiongroup" or something else?

This constant isn't being used currently in this PR, but I added it to better document the `descendant_level` param of `get_descendant_regions()`. Can use it later to translate region level names to ids or vice versa

eyvindn: set_of_regions maybe? depends how mathematically pedantic our users are ;) otherwise "group" is fine too imo.

eyvindn (approved):

api/client/lib.py
nemozen: it's really just  "other" ... not even necessarily a set of other regions.   it could be a water basin for example or other region which is not necessarily expressible as a union of admin regions.

itsjohnward: done

nemozen (approved):

PR #50, opened 2019-04-02
Minor bugfixes

eyvindn:
See commit messages. Please merge directory pr first

### DIFF
--- a/api/client/samples/similar_regions/similar_region.py
+++ b/api/client/samples/similar_regions/similar_region.py
@@ -11,7 +11,7 @@ ACCESS_TOKEN = os.environ['GROAPI_TOKEN']
 
 class SimilarRegion(object):
 
-    def __init__(self, region_properties, regions_to_compare=None):
+    def __init__(self, region_properties, regions_to_compare=None, data_dir=None):
         """
         :param region_properties: A dict containing the properties of regions to use when doing the similarity
         computation. This is by default defined in region_properties.py, but can be adjusted.
@@ -23,7 +23,7 @@ class SimilarRegion(object):
         if not regions_to_compare:
             regions_to_compare = self._regions_avail_for_selection(region_properties)
         self._logger.info("SimilarRegionState: loading...")
-        self.state = SimilarRegionState(region_properties, regions_to_compare, self.client)
+        self.state = SimilarRegionState(region_properties, regions_to_compare, self.client, data_dir=data_dir)
         self._logger.info("SimilarRegionState: done.")
         self._logger.info("BallTree: computing...")
         self.ball = BallTree(self.state.data_standardized, leaf_size=2)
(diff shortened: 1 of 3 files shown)


api/client/samples/similar_regions/similar_region_state.py
```diff
@@ -88,7 +88,7 @@ def load(self):
         self._logger.info("Loading data")
         # Loop through the metric views...
         for name in self.region_properties:
-            path = os.path.join(CACHE_PATH, "{}.nbz".format(name))
+            path = os.path.join(os.path.dirname(__file__), CACHE_PATH, "{}.nbz".format(name))
```
nemozen: We don't want to force the path to be relative to the code base (e.g.  it might not be writeable by the user running the client). Making it relative to the current dir as now is ok, but better yet, how about using the tempfile lib.

api/client/samples/similar_regions/similar_region.py
```diff
@@ -72,6 +72,8 @@ def _similar_to(self, region_id, number_of_regions):
         # list as an index to preserve dimensionality of returned data
         x = self.state.data_standardized[self.state.mapping[region_id], :]
         x = x.reshape(1, -1)
+        assert number_of_regions < self.ball.data.shape[0], "num_regions must be smaller than or equal to total " \
```
nemozen: This would be easier to understand if we make the RHS  self.state.num_regions since that is what the user can more directly control and  so we don't need expose the internals of the balltree.

nemozen: Use the real variable name in the error message (i.e. num_regions or number_of_regions). And let's add the param to the function's docstring.

api/client/samples/similar_regions/similar_region_state.py
eyvindn: tempfile lib in python creates temporary directories/files during the lifetime of the python process. We want these to remain between invocations of similar regions so it's not the best approach. 

Earlier it was relative to the current dir, but then the problem is:
-if the cache directory doesn't exist, should we create it in whatever currdir the user's shell is in? I think that's a bad idea. 
-if the user invokes the file from one or more directories above, the cache directory now changes location. e.g. running 

``` python find_similar_region.py ...```  
vs
``` python similar_regions/find_similar_region.py``` 

will place and look for the cache directory at two different places.

eyvindn: https://docs.python.org/3/distutils/setupscript.html#installing-package-data

This may be useful?

api/client/samples/similar_regions/similar_region.py
eyvindn: the command line param is called num_regions so i'll change this to make it consistent, for error message, variable name and command line param.

PR #51, opened 2019-04-09
Planting intentions model notebook to samples

eyvindn:
planting intentions notebook


eyvindn:
this should look pretty when ReviewNB is added to the repository.

nemozen:
In the text, please link the first mention of entities to their definition e.g. the  "the corn belt" to https://app.gro-intelligence.com/dictionary/regions/100000100 etc.

nemozen:
we should not need to filter out  DC, Hawaii, Alaska, choosing the corn belt does that already

eyvindn:
done x2. 

also realized I uploaded a slightly out of date version, committing a slightly more cleaned up version now.

nemozen (commented):
I think there might a  problem with the crops that you have with "_fsa" e.g. this 
https://app.gro-intelligence.com/dictionary/items/10446 is not corn.

eyvindn:
_fsa item group removed and does a depth first search for items instead.

nemozen:
(diff too large for notebook review tool so I will post comments here)

- "Gro" and "we" are a bit confusing here... Let's use Gro for the api behavior, and reserve "we" for the planting intentions client.

- the log verbosity makes it hard to read this notebook. Let's use debug() for the stuff that's in the loop over each feature, and use info for the top level stuff.

 - How about including a comparison with the following less precise but more robust model
-- make the output variable y be the ratio next_year_acreage/current_year_acreage
--  take county out as a feature, i.e. the goal is to train a single model that works for *all* counties
in other words, we currently have this model

y(t+1, region) - y(t, region) = Model(features(t), region)

and we want to compare it with a simpler model of the form

y(t+1)/y(t) = Model(features(t))

Obviously this is to see if the regional variability is already captured in all the other features like land costs and operating costs etc. Which may or may not be true, but worth seeing.

- instead of outputing each year's scatter and choropleth maps, we should summarize the results in two ways
--  the county-level performance combined across all years (i.e. aggregate the scatter plots into one and the choropleth maps into one )... this gives us the spatial view of the modeling quality
-- a single time series showing the total error (sum of error across counties) as a function of the year, this gives us the temporal view

- replace 1995 with start_year

eyvindn:
got around to addressing the above points. 
ratio-based approach gives similar performance it seems. there's some bias towards 0 in the predictions with "ratio" approach when you look at the scatter plot which I haven't been able to figure out. 

https://app.reviewnb.com/gro-intelligence/api-client/commit/f5e95f279ccb08f15bba7ce4ea1f72cb2324c7a0/

see above^ (n.b.: some output is muddled by reviewnb but it should look pretty in Jupyter)

eyvindn:
total area bug looks to be fixed, was a double counting issue due to expanding crop ids to sub-ids after crop ids were already being expanded by Gro API. ratio based method has slightly lower mean error in acreage.

nemozen:
you can use client.get_descendant_regions(state_id, 5)  to fill in state_to_counties

eyvindn:
> you can use client.get_descendant_regions(state_id, 5) to fill in state_to_counties

going to have to lookup() each county anyway to check it's ERS region so I reckon it's better not to do this twice.

PR #52, opened 2019-04-11
similar regions: add no download option

eyvindn:
--no_download flag uses only data in the cache. prevents attempting to download missing data points every time.

### DIFF
--- a/api/client/samples/similar_regions/similar_region_state.py
+++ b/api/client/samples/similar_regions/similar_region_state.py
@@ -22,7 +22,6 @@ class SimilarRegionState(object):
     def __init__(self, region_properties, regions_to_compare, client, data_dir=None, no_download=False):
         self.client = client
         self._logger = api.client.lib.get_default_logger()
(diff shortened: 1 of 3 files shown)


api/client/samples/similar_regions/similar_region_state.py
```diff
@@ -19,9 +19,10 @@ class SimilarRegionState(object):
     Holds and initializes parameters and provide ssaving/loading logic for a similar_region search.
     """
 
-    def __init__(self, region_properties, regions_to_compare, client, data_dir=None):
+    def __init__(self, region_properties, regions_to_compare, client, data_dir=None, no_download=False):
         self.client = client
         self._logger = api.client.lib.get_default_logger()
+        self.no_download = no_download
```
nemozen: rather than saving it as an instance variable, just use it directly as an argument to load()
besides reducing the state, it also makes sense to have load takea  boolean which controls whether it will download or not

PR #54, opened 2019-04-16
bugfix: similar regions with missing-data regions

eyvindn:
fixes non-sensical results being returned if some of the regions have missing data.

### DIFF
--- a/api/client/samples/similar_regions/similar_region.py
+++ b/api/client/samples/similar_regions/similar_region.py
@@ -55,7 +55,7 @@ class SimilarRegion(object):
                 grandparent = next(self.client.lookup_belongs("regions", parent["id"]), {"name": "", "id": ""})
             else:
                 grandparent = {"name": "", "id": ""}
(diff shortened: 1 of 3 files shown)


eyvindn:
(rebased off development)

api/client/samples/similar_regions/similar_region.py
```diff
@@ -55,7 +55,7 @@ def _format_results(self, sim_regions, region_level_id, dists):
                 grandparent = next(self.client.lookup_belongs("regions", parent["id"]), {"name": "", "id": ""})
             else:
                 grandparent = {"name": "", "id": ""}
-            self._logger.info("{}, {}, {}".format(sim_region_name, parent["name"], grandparent["name"]))
+            self._logger.info("{}, {}, {}".format(sim_region_name.encode('utf8'), parent["name"].encode('utf8'), grandparent["name"].encode('utf8')))
```
nemozen: rather than force an encoding here,  put u in front of the string constant to use the environment's encoding  `u"{}, {}, .."`

eyvindn: done

PR #55, opened 2019-04-16
add soilgrids to similar regions

eyvindn:
branched off some bugfix branches that should be merged first.


api/client/samples/similar_regions/find_similar_regions.py
```diff
@@ -8,18 +8,19 @@
 def main():
     parser = argparse.ArgumentParser(description="Gro Similar Regions")
     parser.add_argument("--region_id", required=True, type=int)
-    parser.add_argument("--csv_output", default=True)
-    parser.add_argument("--region_level", default=None)
+    parser.add_argument("--csv_output", action='store_true')
+    parser.add_argument("--region_level", default=None, type=int)
     parser.add_argument("--number_of_regions", default=10, type=int)
     parser.add_argument("--data_dir", default=None, type=str)
+    parser.add_argument("--no_download", action='store_true')
```
tolothy: as a general comment for the args, i believe it would be helpful, especially to new users, by adding `help=` for key args that explain data types or examples that are excepted.

tolothy: and a comment specific to this line: instead of calling `no_download`, is it more straightforward to have an arg as `download`?

api/client/samples/similar_regions/region_properties.py
```diff
@@ -9,6 +9,7 @@
             'source_id': 43
         },
         "properties": {
+            "type": "timeseries",
```
tolothy: is that information contained in `frequency_id`?

eyvindn: inherently yes - but want to use "type" to denote how to handle the series. so some time series we probably want to do what i do now (fourier transform), but other time series will need different transforms.

I will rename this from "timeseries".

api/client/samples/similar_regions/find_similar_regions.py
eyvindn: doing comment (1). 

as for (2), i'm keeping "no_download" because it's a special case (you have most or all of the data cached and you don't even want similar regions to check if it should download anything else).

eyvindn: (1) done.

api/client/samples/similar_regions/region_properties.py
eyvindn: rename done.

PR #56, opened 2019-04-20
Shapefile support

nemozen:
add region_id -> geojson fn to lib and Client


eyvindn:
tested and lgtm.

eyvindn (approved):
tested & lgtm

PR #57, opened 2019-04-24
Cropmodel dataframe improvements

nemozen:
in get_df()
- convert start_date and end_date columns into datetime type so callers don't have to do it
- use merge dataframes instead of concatenating them

### DIFF
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -56,7 +56,9 @@ class GroClient(Client):
         while self._data_series_queue:
             tmp = pandas.DataFrame(
                 data=self.get_data_points(**self._data_series_queue.pop()))
-            tmp.end_date = pandas.to_datetime(tmp.end_date)
+            if 'end_date' in tmp.columns:
+                tmp.end_date = pandas.to_datetime(tmp.end_date)
+            if 'start_date' in tmp.columns:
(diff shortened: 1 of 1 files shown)


api/client/gro_client.py
```diff
@@ -53,12 +53,15 @@ def __init__(self, api_host, access_token):
...
         """Get the content of all data series in a Pandas frame."""
-        frames = [self._data_frame]
         while self._data_series_queue:
-            frames.append(pandas.DataFrame(data=self.get_data_points(
-                **self._data_series_queue.pop())))
-        if len(frames) > 1:
-            self._data_frame = pandas.concat(frames)
+            tmp = pandas.DataFrame(
+                data=self.get_data_points(**self._data_series_queue.pop()))
+            tmp.end_date = pandas.to_datetime(tmp.end_date)
```
eyvindn: end_date and start_date members won't be available here if no data points are returned.

eyvindn (changes requested):

api/client/gro_client.py
nemozen: Made it conditional.

eyvindn: lgtm

eyvindn (approved):

PR #59, opened 2019-04-25
CLEWS-16588 Add search by source example

itsjohnward:
Showcases search functions in a minimal but useful script in direct response to a frequently asked question.

Also an example of a script-level docstring we can use for other sample scripts, explaining the purpose and linking out to other relevant documentation.


JGrinst (approved):
I like the format and the comments look really helpful to me.

api/client/samples/search.py
```diff
@@ -0,0 +1,45 @@
...
+import os
+from api.client.gro_client import GroClient
+
+API_HOST = 'api.gro-intelligence.com'
+ACCESS_TOKEN = os.environ['GROAPI_TOKEN']
+
+def main():
+    client = GroClient(API_HOST, ACCESS_TOKEN)
+
+    # See what sources match the "Gro" search term
```
nemozen: Using Gro sources is a bit confusing for the first time user.   How about we  use  a more straightforward external source example which has a decent number of data series (but not huge) and is  exotic enough to be interesting, say CASDE or CONAB?

itsjohnward: I wanted to answer the FAQ "what data series does Gro have yield models for?" but the `search_and_lookup()` and `search() -> get_data_series()` examples don't necessarily need to be the same source.

Updated now to show a mix: CASDE along with the Gro Yield Model one. https://github.com/gro-intelligence/api-client/pull/59/commits/686692611942680e06a7593b9d11c88db43247af Helps highlight the fact that Gro sources are treated the same as normal sources. Think that's better?

nemozen: Let's also have an example of searching for the main entities, like items or metrics.  I'm afraid having it named search.py will make people think that source-centric exploration is the right way to approach gro, which is not what we want.

itsjohnward: I have updated this now to showcase:

1. `search()`
2. `search_and_lookup()`
3. `get_data_series()` with item/region
4. `get_data_series()` with source

nemozen (approved):

PR #60, opened 2019-04-30
set point-level multi index on df to allow adding the same series mor…

nemozen:
…e than once without duplicating points


tolothy (approved):
looks good to me

PR #63, opened 2019-05-16
CLEWS 16749 Rename api.client project

itsjohnward:
Rename/deprecation strategy:

1. Leave the old (/api) directory alone and completely copy it to the new location (/gro)
2. Update functions in the old /api lib file to log a warning about the deprecation but otherwise leave it the same. Now it’s locked in time and doesn’t get new updates even if you `git pull` or  `pip install --upgrade` unless you update your config to point to the new package.

Using this script as a test case:

```python
import os

ACCESS_TOKEN=os.environ['GROAPI_TOKEN']

def main():
    for point in client.get_data_points(**{
        'metric_id': 170037,
        'item_id': 270,
        'region_id': 1215,
        'source_id': 32,
        'frequency_id': 9,
        'start_date': '2019-01-01',
        'end_date': '2019-12-31'
    }):
        print(point)

if __name__ == "__main__":
    main()
```

I get this output:

```python

        Deprecation Warning: Please import from the gro package instead of
        api.client.
        api.client will be removed in a future update.
    
{'start_date': '2019-01-01T00:00:00.000Z', 'end_date': '2019-12-31T00:00:00.000Z', 'value': 3.2774030159808514, 'input_unit_id': 61, 'input_unit_scale': 1, 'metric_id': 170037, 'item_id': 270, 'region_id': 1215, 'frequency_id': 9, 'unit_id': 61}
```

But if I change my import from `from api.client.gro_client import GroClient` to `from gro import GroClient` now I get the output:

```python
{'start_date': '2019-01-01T00:00:00.000Z', 'end_date': '2019-12-31T00:00:00.000Z', 'value': 3.2774030159808514, 'input_unit_id': 61, 'input_unit_scale': 1, 'metric_id': 170037, 'item_id': 270, 'region_id': 1215, 'frequency_id': 9, 'unit_id': 61}
```

with no warning.


gro/__init__.py
```diff
@@ -0,0 +1,4 @@
+from gro import lib
+from gro.base_client import Client
+from gro.batch_client import BatchClient
```
eyvindn: Is there a way we can do this without adding BatchClient to __ init __? Reason being that BatchClient depends on third party library (Tornado) which isn't required for normal client, but in this case gets imported nonetheless. 

I suppose the alternative is to rename the files themselves to allow a 'native' import (not sure what to call it, an import without having to use the dot)

eyvindn (changes requested):

gro/__init__.py
itsjohnward: I can put each of these in try/except blocks so if you don't have the dependency it doesn't fail - it just doesn't import that client. Like if you don't have pandas, GroClient can be excluded, and if you don't have tornado BatchClient can be excluded. In that case you could still use Client or lib if you wanted.

Is worth noting that tornado is in the requirements.txt https://github.com/gro-intelligence/api-client/blob/development/requirements.txt#L9 so you should have it already.

Aside from that, I'm not sure how to avoid having that extra redundant `batch_client` layer in `from gro.batch_client import BatchClient`

itsjohnward: ^ I added the exception catching mentioned above, and now, having specifically uninstalled tornado, this is my output:

```shell
$ python quick_start.py

Not able to import BatchClient: No module named 'tornado'
{'start_date': '2019-01-01T00:00:00.000Z', 'end_date': '2019-12-31T00:00:00.000Z', 'value': 3.079994034538774, 'input_unit_id': 61, 'input_unit_scale': 1, 'metric_id': 170037, 'item_id': 270, 'region_id': 1215, 'frequency_id': 9, 'unit_id': 61}
```

eyvindn: > I can put each of these in try/except blocks so if you don't have the dependency it doesn't fail - it just doesn't import that client. Like if you don't have pandas, GroClient can be excluded, and if you don't have tornado BatchClient can be excluded. In that case you could still use Client or lib if you wanted.

OK this works. My original thought was to avoid importing all of the API classes when you import one part of the API. But overhead shouldn't be too bad with these three things. I think this is good for now.

eyvindn: lgtm

api/client/lib.py
```diff
@@ -138,6 +138,11 @@ def get_data(url, headers, params=None, logger=None):
         logger = get_default_logger()
         logger.debug(url)
         logger.debug(params)
+    logger.warning('''
```
eyvindn: Does anyone have anything against moving this to api/\_\_init\_\_.py

That way we get the warning when we import as opposed to just when you run the file. I added the following to the __init__.py file:

```
from client.lib import get_default_logger
logger = get_default_logger()
logger.warning('''
                Deprecation Warning: Please import from the gro package instead of
                        api.client.api.client will be removed in a future update.
''')
```

Seemed to have the intended effect. Thoughts @nemozen @tolothy @itsjohnward 

image

eyvindn (changes requested):
see comment re: warning

api/client/lib.py
eyvindn: also would prevent this warning for triggering every single data call.

itsjohnward: lgtm. Moved

itsjohnward:
Going to recreate this PR with the latest version and remove the warnings for the first version

PR #67, opened 2019-05-29
CLEWS-17697 Update README examples and token instructions

itsjohnward:
1. Added a Gro logo to the top of the README
2. Slimmed down the README to cover only the most common cases and remove superfluous or confusing information (`git clone` alternative installation, pip `--target`, `gro_client` example)
3. Updated the authentication token instructions to reference the new Wiki page: https://github.com/gro-intelligence/api-client/wiki/Authentication-Tokens
4. Added `soybeans.py` and `brazil_soybeans.ipynb` examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+<p align="center"><img width=8% src="https://gro-intelligence.com/images/logo.jpg"></p>
+
 # Gro API Client
 
 https://www.gro-intelligence.com/products/gro-api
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,81 +13,36 @@ Client library for accessing Gro Intelligence's agricultural data platform.
...
+2. Try out [soybeans.py](api/client/samples/crop_models/soybeans.py) to see the crop weighted series feature in action:
 
 ```sh
-gro_client --metric='Production Quantity mass' --item='Corn' --region='United States' --user_email='email@expample.com'
-Password: <enter password when prompted>
-
-or
-
-gro_client --metric='Production Quantity mass' --item='Corn' --region='United States' --token='token-generated-in-setup-steps'
+python crop_modles/soybeans.py
```
nemozen: "modles" -> "models"

README.md
```diff
@@ -11,81 +13,36 @@ Client library for accessing Gro Intelligence's agricultural data platform.
...
-```sh
-cd your-gro-path/api-client/api/client/samples/crop_models/
-python sugar.py
-```
-
-You can also use the Gro CLI as a quick and easy way to request a single data series right on the command line. You can either provide your email and enter a password when prompted, or you can provide your --token to avoid typing a password every time:
+2. Try out [soybeans.py](api/client/samples/crop_models/soybeans.py) to see the crop weighted series feature in action:
 
 ```sh
-gro_client --metric='Production Quantity mass' --item='Corn' --region='United States' --user_email='email@expample.com'
```
nemozen, line 81: we should keep this command line gro_client (or python gro_client.py) example as well.  since it's not udner "samples" maybe we move ti to the beginnig or the end of the examples

also lets use "command line" rather than "cli", which may be obscure for some of  our audience.

nemozen (changes requested):

README.md
itsjohnward: done

README.md
```diff
@@ -11,81 +13,36 @@ Client library for accessing Gro Intelligence's agricultural data platform.
...
-pip install git+https://github.com/gro-intelligence/api-client.git [--target=<your-gro-path>]
-```
-
-If you don't specify your gro installation path, it will be the local [site-packages]( https://stackoverflow.com/questions/31384639/what-is-pythons-site-packages-directory) directory.
-
-### Install with git clone 
-
-```
-git clone https://github.com/gro-intelligence/api-client.git  [<your-gro-path>]
+pip install git+https://github.com/gro-intelligence/api-client.git
```
nadiachil, line 17: you could add a link to installing pip since we aren't giving the alternate option

README.md
```diff
@@ -11,81 +13,36 @@ Client library for accessing Gro Intelligence's agricultural data platform.
...
+If you don't want to enter a password or token each time, you can save the token as an environment variable. Please consult your OS or IDE documentation for information on how to set environment variables, e.g. [setting environment variables in Windows Powershell](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-6) and [Mac OS X/Linux](https://apple.stackexchange.com/questions/106778/how-do-i-set-environment-variables-on-os-x). In some of the sample code, it is assumed that you have the token saved to your environment variables as `GROAPI_TOKEN`.
 
 ## Examples
 
 Navigate to [api/client/samples/](api/client/samples/) folder and try executing the provided examples.
 
-Try [quick_start.py](api/client/samples/quick_start.py)
+1. Try [quick_start.py](api/client/samples/quick_start.py).
```
JGrinst: I think this would be a good opportunity to add context around these samples and why users might find it helpful. E.g. "This quick start file demonstrations some helpful frameworks for engaging with the Gro API Client. The indicates the entities of interest, uses the get_data_series function to return all related series, sets a time range restriction, then retrieves get_data_points for all series, and finally outputs the request into a CSV"

nadiachil (commented):
this looks really nice - one idea to make more informative is just a bit more information surrounding the samples file and how we suggest users familiarize themselves with requests through the samples/ more info about the data you are requesting through soybeans.py instead of "in action" like describe how it will create a new file where you can look at the data format etc.

README.md
JGrinst: If we add the command line back, I think more instruction should be provided behind what it's doing and why it might be helpful. I was trying to understand what this sample was doing yesterday without much luck. I believe it returns a random example? If so, I think we should use an example that is more helpful than that.

README.md
itsjohnward: pip is in the `Prerequisites` section already, which I did not change

And besides, `pip` is included in Python by default in Python >=2.7.9 & Python 3 >=3.4 (see https://pip.pypa.io/en/stable/installing/). We enforce a minimum version of 2.7.6 in https://github.com/gro-intelligence/api-client/blob/development/setup.py#L18 though we haven't targeted a minimum 3.x version yet. So, it is possible they might not have pip, but it should be a small cross-section

README.md
itsjohnward: I just added this back with some description.

I want to minimize it since it's not what we want to encourage for use in production. Added a note: "This tool is useful for simple queries, but anything more complex should be done using the Python packages."

Let me know if you think that covers it

README.md
itsjohnward: Added. I want to put more detailed info in the quick_start file itself as in https://github.com/gro-intelligence/api-client/pull/59/files#diff-39a3741db0aea5fdfb9fdd7bb405c162R1 so don't want to duplicate too much, but I added some of the most important info on what to expect from it

JGrinst (approved):

nemozen (approved):

PR #71, opened 2019-06-13
Add Analysis Kits Directory

stephenbernardi:
Create a new directory for "analysis kits" and add stocks to use price model


api/client/samples/analysis_kits/stocks_to_use_price_model.py
```diff
@@ -0,0 +1,271 @@
+"""
+Sample Gro API Client script for regressing stocks-to-use ratios vs price
+
+This script creates a basic CME price valuation model using the US stocks-to-use ratio.
```
JGrinst: Could you add a bit more descriptive information in the comments? Perhaps a sentence or two describing the method and then an indication of what the output of this will be.

stephenbernardi: added a few sentences, PTAL

api/client/samples/analysis_kits/stocks_to_use_price_model.py
```diff
@@ -0,0 +1,287 @@
...
+    contract_month : integer
+
+    Returns
+    -------
+    pandas.Series
+    """
+
+    market = ref[crop.lower()]['id']
+    
+    client = GroClient(API_HOST, TOKEN)
```
nemozen: you already have a client object defined  one line 34 don't need to make another one here, just keep using that one.

api/client/samples/analysis_kits/stocks_to_use_price_model.py
```diff
@@ -0,0 +1,287 @@
...
+        Gro region id
+    Returns
+    -------
+    pandas.Series
+    """
+    
+    c_ref   = ref[crop.lower()]
+    crop_id = c_ref['id']
+
+    client = GroClient(API_HOST, TOKEN)
```
nemozen: ditto, just reuse the client object you have define globally

api/client/samples/analysis_kits/stocks_to_use_price_model.py
```diff
@@ -0,0 +1,287 @@
...
+    region : int
+        Gro region id
+    Returns
+    -------
+    pandas.Series
+    """
+    c_ref   = ref[crop.lower()]
+    crop_id = c_ref['id']
+    
+    client = GroClient(API_HOST, TOKEN)
```
nemozen: ditto reuse client

api/client/samples/analysis_kits/stocks_to_use_price_model.py
```diff
@@ -0,0 +1,287 @@
...
+
+    Parameters
+    ----------
+    crop : string
+        'corn' or 'soybeans'
+    Returns
+    -------
+    pandas.Series
+    """
+    united_states = 1215
```
nemozen: this constant is defined in a few different places, here, in the default arg and in the first function... it would be clearer to just define it once and for all as a global constant

UNITED_STATES_REGION_ID = 1215 

and use that throughout, and remove the default arg value in the get_consumption and get_ending_stocks functions

api/client/samples/analysis_kits/stocks_to_use_price_model.py
```diff
@@ -0,0 +1,287 @@
...
+
+from sklearn import linear_model
+
+from api.client.gro_client import GroClient
+
+API_HOST = 'api.gro-intelligence.com'
+TOKEN    = os.environ['GROAPI_TOKEN']
+client   = GroClient(API_HOST, TOKEN)
+
+ref = {'corn':{'id':274, 'futures_contract_month':12},
```
nemozen: rather than hard-coding the item ids, better to use 

client.search_for_entity('items', 'corn')
client.search_for_entity('items', 'soybeans')

nemozen (changes requested):

api/client/samples/analysis_kits/stocks_to_use_price_model.py
```diff
@@ -0,0 +1,287 @@
...
+(December futures for corn and November futures for soybeans) price for the 
+relevant report month. For example, the corn-June scatterplot will compare the US
+stocks-to-use ratio from all June reports relative to the corresponding monthly 
+average December corn futures price.
+
+The output of this script will be a PDF file for each modeled crop, saved in
+the current working directory. Each PDF will have 13 pages in total, 
+including a scatterplot for each report month as well as a 
+scatterplot summarizing the entire sample.
+
```
nemozen, line 19: can you  add a link to gro public display with the 3 data series used as inputs here

JGrinst (approved):

api/client/samples/analysis_kits/stocks_to_use_price_model.py
stephenbernardi: added in 1st paragraph

api/client/samples/analysis_kits/stocks_to_use_price_model.py
stephenbernardi: updated to use a global client object. I added "client.get_df().copy()" because get_df was throwing an error after I converted reporting_date column to datetime and tried to add another series.

ValueError: You are trying to merge on datetime64[ns] and object columns. If you wish to proceed you should use pd.concat

nemozen: ok now the GroClient.get_df() now returns a df with reporting_date col cast to datetime

nemozen: cleaned up this code to use reporting_date from client's get_df

nemozen (approved):

PR #77, opened 2019-07-11
remove unecessary assert (causes py3 error)

eyvindn:
dict_values is no longer subscriptable in py3 so this assert fails. however we don't need it anymore.


tolothy:
@eyvindn could you explain more why we no longer need the check?

eyvindn:
the check is no longer needed because it was a sanity check during model development and at this point we're sure it's the correct unit.

nemozen (approved):

PR #78, opened 2019-07-11
optionally limit number of results for search_and_lookup

nemozen:
Search may return a very large number of results, and performing
lookup on each one can be time consuming, so we optionally limit the
number of desired results


itsjohnward (approved):
LGTM

PR #79, opened 2019-07-17
Fix Stocks to Use Price Model Sorting Error 

stephenbernardi:
This was causing the script to fail in python 3


JGrinst (approved):
Works for me now, and looks good!

PR #80, opened 2019-07-19
refactor prevented plant

eyvindn:
Refactor prevented plant to be simpler by using more pandas functions instead of hand-rolled functions and move some things into an external helper dependency. Allows soy + corn model to be separate as well.


itsjohnward:
Closing this PR as stale.

PR #82, opened 2019-07-22
remove USA_STATES  constant

nemozen:


api/client/samples/similar_regions/README.md
```diff
@@ -14,18 +14,32 @@ For a general technical overview of the approach taken, please take a read throu
...
+
+Now suppose we wanted to expand our search to all provinces in the world. In that case, we just omit regions_to_compare, but we still leave the region level as `4`.  Constructing the SimilarRegion object will now take even longer since the model is getting data for the whole world.  
+
+```
+>>> sim = SimilarRegion(region_properties)
+>>> for result in sim.similar_to(WISCONSIN_REGION_ID, 10, 4):
+        print(result)
+```
+
+If we wanted to compare it to countries instead of provinces we would change the region_level from `4` to `3`, and for  districts we would use region level `5`.
```
itsjohnward: Could import and use the `lib.REGION_LEVELS` constant instead of explaining the numbers

PR #85, opened 2019-07-31
CLEWS-19599 Client-side Unit Conversion

itsjohnward:
Changes from add_unit_conversion branch:

1. Removed the `convert_unit` function from lib and Client and placed it instead in `GroClient`.
2. `convert_unit` takes a point and a target unit, rather than a value, a from_unit, and a to_unit. A modified `point` dict is returned instead of a single value.
3. If the original unit and the target unit are identical, exit early
4. Extended the GroClient's `get_data_points()` function to first call the `Client` version and then apply unit conversion automatically if a target unit is provided.
5. Added memoization to lib functions where applicable. This eliminates the multiple redundant lookups of unit conversion factors when looping through a long list of get_data_points responses

Tested with the following script:

```py
import os

def main():

    YIELD=170037
    CORN=274
    UNITED_STATES=1215
    GRO_YM=32
    ANNUAL=9
    TONNES_PER_HECTARE=61
    BUSHELS_PER_ACRE=284

    print('No conversion')
        metric_id=YIELD,
        item_id=CORN,
        region_id=UNITED_STATES,
        source_id=GRO_YM,
        frequency_id=ANNUAL,
        unit_id=TONNES_PER_HECTARE
    ):
        print(point['value'], client.lookup_unit_abbreviation(point['unit_id']))

    print('Convert into identical unit')
        metric_id=YIELD,
        item_id=CORN,
        region_id=UNITED_STATES,
        source_id=GRO_YM,
        frequency_id=ANNUAL,
        unit_id=TONNES_PER_HECTARE
    ):
        print(point['value'], client.lookup_unit_abbreviation(point['unit_id']))

    print('Convert into different unit')
        metric_id=YIELD,
        item_id=CORN,
        region_id=UNITED_STATES,
        source_id=GRO_YM,
        frequency_id=ANNUAL,
        unit_id=BUSHELS_PER_ACRE
    ):
        print(point['value'], client.lookup_unit_abbreviation(point['unit_id']))

    print('Convert units when there\'s no data:', list(client.get_data_points(
        metric_id=YIELD,
        item_id=CORN,
        region_id=UNITED_STATES,
        source_id=GRO_YM,
        frequency_id=ANNUAL,
        start_date='2020-01-01',
        unit_id=BUSHELS_PER_ACRE
    )))

if __name__ == "__main__":
    main()
```

Output:

```
$ python CLEWS-19599.py 
No conversion
10.825978048472363 t/ha
10.957399391107465 t/ha
11.179951735272198 t/ha
10.053230574702063 t/ha
Convert into identical unit
10.825978048472363 t/ha
10.957399391107465 t/ha
11.179951735272198 t/ha
10.053230574702063 t/ha
Convert into different unit
172.475851555 bu/acre
174.569612311026 bu/acre
178.115241620776 bu/acre
160.164697959569 bu/acre
Convert units when there's no data: []
```

### DIFF
--- a/api/client/__init__.py
+++ b/api/client/__init__.py
@@ -6,8 +6,6 @@ class Client(object):
 
     """API client with stateful authentication for lib functions. """
 
-    __unit_names = {}
-
     def __init__(self, api_host, access_token):
         self.api_host = api_host
         self.access_token = access_token
@@ -22,10 +20,7 @@ class Client(object):
         return lib.lookup(self.access_token, self.api_host, entity_type, entity_id)
 
     def lookup_unit_abbreviation(self, unit_id):
-        """Wrapper to lookup unit names, with local cache to avoid repeated lookups."""
-        if unit_id not in self.__unit_names:
-            self.__unit_names[unit_id] = self.lookup('units', unit_id)['abbreviation']
-        return self.__unit_names[unit_id]
+        return self.lookup('units', unit_id)['abbreviation']
 
     def get_data_series(self, **selection):
         return lib.get_data_series(self.access_token, self.api_host, **selection)
@@ -58,6 +53,3 @@ class Client(object):
     def get_descendant_regions(self, region_id, descendant_level=None):
         return lib.get_descendant_regions(self.access_token, self.api_host,
                                           region_id, descendant_level)
-
-    def convert_unit(self, value, from_unit_id, to_unit_id):
-        return lib.convert_unit(self.access_token, self.api_host, value, from_unit_id, to_unit_id)
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -76,8 +76,7 @@ class GroClient(Client):
                 tmp.reporting_date = pandas.to_datetime(tmp.reporting_date)
             if self._data_frame is None:
                 self._data_frame = tmp
-                self._data_frame.set_index(
-                    [col for col in DATA_POINTS_UNIQUE_COLS if col in tmp.columns])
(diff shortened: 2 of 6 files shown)


tolothy:
@itsjohnward 

1. Could you also test a case when the unit is not convertible?
2. I think there is one think i overlooked when implementing `convert_unit`, is to assert the base_unit of from_unit and to_unit are the same. could you add that functionality please, to prevent people from converting `meter` to `bushel` for example.

itsjohnward:
> @itsjohnward
> 
> 1. Could you also test a case when the unit is not convertible?
> 2. I think there is one think i overlooked when implementing `convert_unit`, is to assert the base_unit of from_unit and to_unit are the same. could you add that functionality please, to prevent people from converting `meter` to `bushel` for example.

For 1, working on adding the test cases into pytest. Will add a non-convertible case
For 2, we need additional info since `lookup('units')` doesn't tell you what the base unit is. Need to expose a new endpoint to the API client in order to support that. And similarly, it's tough currently to discover what units are available to use. You just have to know the id up-front. So, a new endpoint's needed for full unit conversion support anyway

Also I just realized the lru_cache decorator I used was added in 3.2 and isn't available in 2.7, so I'll replace that with a 2.7-backported one.

3 TODOs

tolothy:
sounds good. maybe you can do 1 and 3 for now and leave 2 as a future TODO so we release this as soon as possible?

itsjohnward:
> sounds good. maybe you can do 1 and 3 for now and leave 2 as a future TODO so we release this as soon as possible?

Will do

tolothy (approved):

PR #98, opened 2019-08-14
CLEWS-15452 At-time query example

itsjohnward:
Primary changes include:
1. Adding at_time as an input for get_data_points
2. The Jupyter notebook explaining the use cases for the at-time query feature

Secondary changes:
1. For functions in the lib that take a variable set of keyword arguments, I added docstrings for the expected keywords instead. This way, I could add each argument's expected type and some extra description. Explaining the at_time argument in particular and linking to the new example notebook.
2. I was getting random failures on doctest since the dicts aren't guaranteed to have keys in the same order every time. So I adjusted the doctests to test equality rather than for the exact output. See the doctest warnings section here: https://docs.python.org/2/library/doctest.html#warnings
3. Previous docstring for `lookup_belongs()` erroneously included `'sources'` in the list of valid types. Removed.
4. `lookup()` didn't have `'frequencies'` listed as a valid type. Added.


api/client/samples/at-time-query-examples.ipynb
```diff
@@ -0,0 +1,407 @@
+{
```
JGrinst, line 1: <p>"If a source&nbsp;is reporting a "total exports" number for the January 1st to December 31st..."</p><p> </p><p>The "the" before January is not needed.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/98/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

JGrinst (approved):
I haven't tested the run of the notebook, but it's very well put together.

api/client/samples/at-time-query-examples.ipynb
itsjohnward: removed

tolothy, line 1: <p>It might be worth pointing out that "forecast" data points usually are available prior to the <code>end_date</code>, to distinguish it from the previous example of LST. Besides, their source_lag might carry no information. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/98/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

itsjohnward: Added

tolothy (approved):

PR #106, opened 2019-09-12
CLEWS-20681 Add functions to get region_ids given latitude and longitude

jeanzhaoyi:
Tested with 
`client.get_region_info_given_lat_lon([(30,120)], output_region_level=4)`
And got [13348], or Zhejiang province.

Also tested with 
` client.get_region_info_given_lat_lon([(30,120), (41, -74)], output_region_level=5)`
And got [141649 138529], or "Hangzhou" city and "Bergen" county respectively.


api/client/lib.py
```diff
@@ -844,6 +848,95 @@ def get_geojson(access_token, api_host, region_id):
...
+                      'coordinates': [[[[-38.394, -4.225], ...]]]}, ...]}, ...]
+    """
+    region_id_str = '&'.join(['regionIds=' + str(id_) for id_ in region_id_list])
+    url = '/'.join(['https:', '', api_host, 'v2/geocentres?includeGeojson=True&' + region_id_str])
+    headers = {'authorization': 'Bearer ' + access_token}
+    resp = get_data(url, headers)
+    geometry_list = [json.loads(object_['geojson'])['geometries'][0] for object_ in resp.json()['data']]
+    return geometry_list
+
+def get_region_id_given_lat_lon(access_token, api_host, latitude, longitude, output_region_level):
```
itsjohnward: This is derivative behavior / shouldn't be in the lib. See the note at the top of this file: "Helper functions or shims or derivative functionality should appear in the client classes rather than here."

The lib should just be wrappers around API endpoints and no other logic.

api/client/lib.py
```diff
@@ -844,6 +848,95 @@ def get_geojson(access_token, api_host, region_id):
...
+    access_token : string
+    api_host : string
+    region_id_list : integer list
+
+    Returns
+    -------
+    [{'type': 'MultiPolygon',
+                      'coordinates': [[[[-38.394, -4.225], ...]]]}, ...]}, ...]
+    """
+    region_id_str = '&'.join(['regionIds=' + str(id_) for id_ in region_id_list])
```
itsjohnward: This url construction `?regionIds=1&regionIds=2&regionIds=3` etc. pretty easily starts running into character limits. You can fit many more using the format `?regionIds=[1,2,3]`

api/client/lib.py
```diff
@@ -844,6 +848,95 @@ def get_geojson(access_token, api_host, region_id):
         return json.loads(region['geojson'])
     return None
 
+@memoize(maxsize=None)
+def get_geometry_list(access_token, api_host, region_id_list):
```
itsjohnward: I don't think this should be its own function - I think I want to just update `get_geojson()` to take either an int or a list of ints.

api/client/lib.py
```diff
@@ -844,6 +848,95 @@ def get_geojson(access_token, api_host, region_id):
...
+    headers = {'authorization': 'Bearer ' + access_token}
+    resp = get_data(url, headers)
+    geometry_list = [json.loads(object_['geojson'])['geometries'][0] for object_ in resp.json()['data']]
+    return geometry_list
+
+def get_region_id_given_lat_lon(access_token, api_host, latitude, longitude, output_region_level):
+"""Look up details of regions of the given level contained by a region.
+
+    Given any region by id, recursively get all the descendant regions
+    that are of the specified level.
```
itsjohnward: Comment copied from get_descendant_regions?

api/client/lib.py
```diff
@@ -12,6 +12,9 @@
 import logging
 import requests
 import time
+import pandas as pd
+import geopandas as gpd
+from shapely.geometry import shape, Point
```
itsjohnward: Shouldn't need any of these dependencies in the lib

api/client/lib.py
itsjohnward: Output doesn't match method name

With this name, I would expect to get an id, not a Geopandas Spatial Point data frame

itsjohnward (changes requested):

api/client/lib.py
jeanzhaoyi: Moved to gro_client.py

api/client/lib.py
jeanzhaoyi: Done

api/client/lib.py
jeanzhaoyi: Done

api/client/lib.py
jeanzhaoyi: Fixed

api/client/lib.py
jeanzhaoyi: Moved to the gro_client.py

api/client/lib.py
jeanzhaoyi: changed to `get_region_info_given_lat_lon`

api/client/gro_client.py
```diff
@@ -237,6 +238,67 @@ def convert_unit(self, point, target_unit_id):
         point['unit_id'] = target_unit_id
         return point
 
+    def get_region_info_given_lat_lon(self, latitude, longitude, output_region_level):
```
nemozen: i think it makes more sense to make the input a list of (lat, lon) pairs rathre than two lists.

api/client/gro_client.py
```diff
@@ -237,6 +238,67 @@ def convert_unit(self, point, target_unit_id):
...
+        Parameters
+        ----------
+        latitude : list of float
+        longitude: list of float
+        output_region_level : integer
+            The region level of interest. See REGION_LEVELS constant.
+
+        Returns
+        -------
+        Geopandas Spatial Point data frame with column names of region ids
```
nemozen: just return a region id  (since there's already a way to get shape file from region id)

api/client/gro_client.py
```diff
@@ -237,6 +238,67 @@ def convert_unit(self, point, target_unit_id):
...
+    """
+        points_geometry = [Point(xy) for xy in zip(longitude, latitude)]
+        gdf = gpd.GeoDataFrame(
+            pandas.DataFrame({'latitude': latitude, 'longitude': longitude}),
+            geometry=points_geometry, crs={'init': u'epsg:4326'})
+
+        continent_ids = range(11, 18)
```
nemozen: rather than implementing each level, let's make this a loop from some tarting region (by default the whole world) going down to the next levels until we reach  output_region_level.

so it would look something like

```
def get_region_info_given_lat_lon(self, lat, lon, output_region_level, current_region_id=1):
 """ Find descendant of current_region_id, at the output_region_level, 
       that contains the given lat, lon point"""

    current_region_level = lookup('regions', current_region_id)['level']
    if output_region_level == current_region_level:
          return parent_region_id
     region_ids = get_descendant_regions(parent_region_id, current_region_level + 1)
     for region_id in region_ids:
                call get_geojson to get polygon for that region_id
                if lat lon is inside polygon: ## <- this is where we do the actual geometry
                      return get_region_info_given_lat_lon(lat, lon, current_region_level + 1)
      return None
```

api/client/gro_client.py
```diff
@@ -237,6 +238,67 @@ def convert_unit(self, point, target_unit_id):
...
+        continents = gpd.GeoDataFrame(pandas.DataFrame({
+                       'continent_id': continent_ids, 
+                       'geometry': [shape(g['geometries'][0]) for g in lib.get_geojson(acess_token, api_host, continent_ids)]
```
nemozen: Don't call lib since you are inside GroClient. Use self.get_geojson(ids)

nemozen (changes requested):

api/client/gro_client.py
jeanzhaoyi: Done

api/client/gro_client.py
jeanzhaoyi: Did change the code into recursive format but modified a bit to allow for the scenario of having multiple parent_region_ids

api/client/gro_client.py
jeanzhaoyi: fixed

api/client/gro_client.py
jeanzhaoyi: still returning the geopandas data frame because we need to keep the reference of which lat lon point refers to which region_id

nemozen: Since this is a fairly basic function, like get_descendant_regions, search and lookups etc., it's probably better to only use basic list rather than  dataframes in the input or output.  input a list of lat, lon pairs, and return a list of region ids.

jeanzhaoyi: Done

api/client/gro_client.py
```diff
@@ -6,17 +6,18 @@
...
 import itertools
 import functools
+import math
 import os
 import pandas
 import sys
 import unicodecsv
 from api.client import cfg, lib, Client
-
+import geopandas as gpd
```
itsjohnward, line 15: Shippable's failing because we don't have geopandas and shapely in our requirements.

And actually including this requirement is a big deal since geopandas doesn't work on Windows pip and we *require* users to use pip to install the client. So adding this effectively breaks all Windows support as is.

Will need to separate this out into an additional geo-utils library or something that is not installed by default, or start distributing binaries on Anaconda Cloud so conda can be our first-class Windows installer

cc @JGrinst

itsjohnward (changes requested):

api/client/gro_client.py
nemozen: yes, we'd like to avoid geopandas as a requirement... I think we can get the functionalitywe want via shapely but that might have the same underlying issue

PR #112, opened 2019-10-04
CLEWS-21548 check if point's value is None before converting

alexsherman:
None is None in any unit system

### DIFF
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -212,7 +212,7 @@ class GroClient(Client):
             new unit_id. Other properties are unchanged.
 
         """
-        if point.get('unit_id') is None or point.get('unit_id') == target_unit_id:
+        if point['value'] is None or point.get('unit_id') is None or point.get('unit_id') == target_unit_id:
             return point
         from_convert_factor = self.lookup(
             'units', point['unit_id']
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -212,7 +212,7 @@ class GroClient(Client):
             new unit_id. Other properties are unchanged.
 
         """
-        if point['value'] is None or point.get('unit_id') is None or point.get('unit_id') == target_unit_id:
+        if point.get('unit_id') is None or point.get('unit_id') == target_unit_id:
             return point
         from_convert_factor = self.lookup(
             'units', point['unit_id']
@@ -221,9 +221,6 @@ class GroClient(Client):
             raise Exception(
                 'unit_id {} is not convertible'.format(point['unit_id'])
             )
-        value_in_base_unit = (
-            point['value'] * from_convert_factor.get('factor')
-         ) + from_convert_factor.get('offset', 0)
         to_convert_factor = self.lookup(
             'units', target_unit_id
         ).get('baseConvFactor')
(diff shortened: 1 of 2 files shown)


api/client/gro_client.py
```diff
@@ -212,7 +212,7 @@ def convert_unit(self, point, target_unit_id):
             new unit_id. Other properties are unchanged.
 
         """
-        if point.get('unit_id') is None or point.get('unit_id') == target_unit_id:
+        if point['value'] is None or point.get('unit_id') is None or point.get('unit_id') == target_unit_id:
```
itsjohnward: I think we should still swap the `point['unit_id']` value to the new unit id, if it passes the non-convertible checks.

A lot of people are going to be using this output in pandas data frames and pivoting on these columns, so if we leave it as the old unit_id, the Nones would get split off from the rest of the data if the non-None values got their ids updated but these didn't.

itsjohnward: Can do something like

```py
if point['value'] is not None:
    value_in_base_unit = (
        point['value'] * from_convert_factor.get('factor')
    ) + from_convert_factor.get('offset', 0)
    point['value'] = float(
        value_in_base_unit - to_convert_factor.get('offset', 0)
    ) / to_convert_factor.get('factor')
point['unit_id'] = target_unit_id
```

I still want those `if not *.get('factor'):` Exceptions to come before return though, for the same reason as the above ^. If one point's getting converted, they should all get converted

itsjohnward (changes requested):
please add two unit tests to gro_client_test.py:

```py
assert client.convert_unit({ 'value': None, 'unit_id': 37 }, 36) == {
  'value': None,
  'unit_id': 36
}
```

and

```py
with pytest.raises(Exception):
  assert client.convert_unit({ 'value': None, 'unit_id': 10 }, 43)
```

I'd expect both of those to work

api/client/gro_client.py
itsjohnward: As noted in shippable https://app.shippable.com/github/gro-intelligence/api-client/runs/714/1/tests this is failing on `assert client.convert_unit({}, 36) == {}`

This happens when there's no data. Instead of `[]` like you might expect, you get `[{}]` so you can't assume there's a 'value' property in that dict. Have to use `point.get('value') is None` instead

itsjohnward (changes requested):

api/client/gro_client.py
alexsherman: Gotcha, will change

alexsherman: Okay, I'll do something like that, agree that all of the points should end up in the same unit even if None

itsjohnward (approved):

itsjohnward (approved):
LGTM. Tested with

```py
import os

for point in client.get_data_points(**{'metric_id': 2100031, 'item_id': 2039, 'region_id': 139802, 'source_id': 35, 'frequency_id': 1}):
    print(point)

for point in client.get_data_points(**{'metric_id': 2100031, 'item_id': 2039, 'region_id': 139802, 'source_id': 35, 'frequency_id': 1, 'unit_id': 8}):
    print(point)
```

I'll go ahead and merge this one, and then follow it up with a version increment PR so updating users can get the fix.

PR #114, opened 2019-10-17
make a print statement compatible with python 3

tolothy:


itsjohnward (approved):

itsjohnward:
Just ran through the whole notebook on this branch in python 3 and had no trouble. Looks like that was the only spot. Merging

PR #116, opened 2019-10-23
CLEWS-20278 Add list_of_series compatible formatting function

itsjohnward:
Using the following test case:

```
        metric_id=3290001,
        item_id=95,
        region_id=[1215,1037],
        source_id=25,
        frequency_id=9
    ):
        print(point)
```

Outputs points like:

```
{
    'start_date': '2008-01-01T00:00:00.000Z',
    'end_date': '2008-12-31T00:00:00.000Z',
    'value': 125500,
    'metric_id': 3290001,
    'item_id': 95,
    'region_id': 13101,
    'partner_region_id': 0,
    'frequency_id': 9,
    'source_id': 25,
    'unit_id': 41,
    'belongs_to': {
        'frequencyId': 9,
        'itemId': 95,
        'metricId': 3290001,
        'partnerRegionId': 0,
        'regionId': 1215,
        'sourceId': 25
    }
}
```

Note that the region_id requested, 1215, is in the belongs_to property of the output point, but the point's region_id is 13101, meaning that 1215 got split up into its children and this is one of them.

The additional belongs_to property is the only difference between the output of the old get_data_points() and this new one.

THIS IS AN UNSTABLE INTERFACE. Note that the `"expansionType"` needed to behave in this manner will likely change before this feature is launched. The input/output format is expected to be representative of the final output though.


api/client/lib.py
```diff
@@ -613,6 +623,30 @@ def get_crop_calendar_data_points(access_token, api_host, **selection):
...
+                        'start_date': point[0],
+                        'end_date': point[1],
+                        'value': point[2],
+                        'metric_id': series['series']['metricId'],
+                        'item_id': series['series']['itemId'],
+                        'region_id': series['series']['regionId'],
+                        'partner_region_id': series['series'].get('partnerRegionId', 0),
+                        'frequency_id': series['series']['frequencyId'],
+                        'source_id': series['series']['sourceId'],
+                        'unit_id': series['series']['inputUnitId'],
```
itsjohnward: TODO: make unit_id available instead of input_unit_id

itsjohnward: done

itsjohnward:
Closing in favor of https://github.com/gro-intelligence/api-client/pull/141

PR #93, opened 2019-08-08
J grinst patch 3

JGrinst:
I'm proposing here to create a new folder - within the crop models folder - for collecting all of the needed information for the Brazil Soybeans model.

There should be:
README
brazil_soybeans.ipynb
requirements.txt
__init__

PR #118, opened 2019-10-25
simplify samples/crop_models and README organization

itsjohnward:
Moved in https://github.com/gro-intelligence/api-client/pull/93 into its own directory. Expect to see it reorganized back up into the main crop_models folder in the future, but as a more immediate fix, point this link at the Brazil soybeans directory


README.md
```diff
@@ -47,7 +47,7 @@ If the API client is installed and your authentication token is set, a csv file
 python crop_models/soybeans.py
 ```
 
-3. See [brazil_soybeans.ipynb](https://github.com/gro-intelligence/api-client/blob/development/api/client/samples/crop_models/brazil_soybeans.ipynb) for a longer, more detailed demonstration of many of the API's capabilities in the form of a Jupyter notebook.
```
nemozen: let's do the reverse as discussed

JGrinst:
As discussed, I don't think this is the reorganization we want. We should not force requirements on models that do not need them. Brazil TRS model does not need geopandas and we shouldn't make users think it's a requirement, since windows users would then also need anaconda to run it.

If we want to have a single requirements doc for crop models, that's fine, but then Brazil TRS shouldn't be a crop model. Let's bring it up to it's own folder.

nemozen:
ok updated and tested both examples

nemozen (approved):

nemozen:
> As discussed, I don't think this is the reorganization we want. We should not force requirements on models that do not need them. Brazil TRS model does not need geopandas and we shouldn't make users think it's a requirement, since windows users would then also need anaconda to run it.
> 
> If we want to have a single requirements doc for crop models, that's fine, but then Brazil TRS shouldn't be a crop model. Let's bring it up to it's own folder.


Yup... turns out the only diff is seaborn which is a simple viz package and statsmodel where we use one regression model which overlaps with sklearn. Neither one is particularly onerous but we can clean that up and reduce the deps a bit separately.

api/client/samples/crop_models/requirements.txt
```diff
@@ -1,7 +1,11 @@
-pandas
-matplotlib
+descartes
```
tolothy, line 1: descartes is a dependency of shapely, which itself is a dependency of geopandas. does that mean we dont need to list it in requirements explicitly?

api/client/samples/crop_models/requirements.txt
```diff
@@ -1,7 +1,11 @@
...
 geopandas
-scikit-learn
 ipython
 jupyter
```
tolothy, line 4: do we only need one of the two?

api/client/samples/crop_models/requirements.txt
nemozen: good question... I think it was added in the TRS notebook since that didn't need geopandas. Doesn't hurt to leave be more explicit here.

nemozen:
> separately



> > As discussed, I don't think this is the reorganization we want. We should not force requirements on models that do not need them. Brazil TRS model does not need geopandas and we shouldn't make users think it's a requirement, since windows users would then also need anaconda to run it.
> > If we want to have a single requirements doc for crop models, that's fine, but then Brazil TRS shouldn't be a crop model. Let's bring it up to it's own folder.
> 
> Yup... turns out the only diff is seaborn which is a simple viz package and statsmodel where we use one regression model which overlaps with sklearn. Neither one is particularly onerous but we can clean that up and reduce the deps a bit separately.

We tested `conda install geopandas` from scratch on  windows and it worked.  As discussed, I have changed the main readme to say anaconda is recommended for windows.  

overall, we don't want to overwhelm the novice user with too many ifs in the readmes, we just want to pick the most pain free most general installation and recommend that. ( ditto for pip install vs git clone for example).

api/client/samples/crop_models/requirements.txt
nemozen: ditto

nemozen:
> > separately
> 
> > > As discussed, I don't think this is the reorganization we want. We should not force requirements on models that do not need them. Brazil TRS model does not need geopandas and we shouldn't make users think it's a requirement, since windows users would then also need anaconda to run it.
> > > If we want to have a single requirements doc for crop models, that's fine, but then Brazil TRS shouldn't be a crop model. Let's bring it up to it's own folder.
> > 
> > 
> > Yup... turns out the only diff is seaborn which is a simple viz package and statsmodel where we use one regression model which overlaps with sklearn. Neither one is particularly onerous but we can clean that up and reduce the deps a bit separately.
> 
> We tested `conda install geopandas` from scratch on windows and it worked. As discussed, I have changed the main readme to say anaconda is recommended for windows.
> 
> overall, we don't want to overwhelm the novice user with too many ifs in the readmes, we just want to pick the most pain free most general installation and recommend that. ( ditto for pip install vs git clone for example).

cleaned up the same issue in ../prevented_plant/README.md  Now they all reference the common setup instructions, shouldbe much easier.   interestingly in that dir geopandas in windows is assumed to just work. ( But there's as separate issue with xgboost on windows, which is handled nicely in the code with a fallback alg).

api/client/samples/crop_models/requirements.txt
```diff
@@ -1,7 +1,11 @@
...
-descartes
+matplotlib
+numpy
+pandas
+scikit-learn
+seaborn
+sklearn
```
itsjohnward: Remove sklearn

On PyPI it exists as an old version https://pypi.org/project/sklearn/, but on Anaconda Cloud it doesn't exist at all and breaks the install. `scikit-learn` is the right one on both

api/client/samples/crop_models/README.md
```diff
@@ -0,0 +1,7 @@
+# Gro API Client examples: Crop Models
+
+1. Follow the [instructions to install the Gro API Client](../../../../README.md)
+1. Download the [requirements file in this directory](./requirements.txt)
+2. Run `pip install -r requirements.txt` (for MacOS and Linux) or `conda install --file requirements.txt` (for Windows) to automatically install the required dependencies
+3. Download the notebook to your working directory
```
itsjohnward: "notebook or script"? There are some python scripts mixed in here too - not all notebooks

And similarly on the next line, run jupyter notebook OR python script.py?

itsjohnward (commented):
Tested all the models from a fresh anaconda environment, and after removing sklearn from the requirements file they all work

api/client/samples/crop_models/requirements.txt
nemozen: done

api/client/samples/crop_models/README.md
nemozen: done

itsjohnward:
https://github.com/gro-intelligence/api-client/tree/e010553e35e37e330abbe8311107407c7f2f2b63/api/client/samples/crop_models

The brazil_soybeans folder is still there, with brazil_soybeans_model_overview.md

Want to remove that file and directory?

nemozen:
> https://github.com/gro-intelligence/api-client/tree/e010553e35e37e330abbe8311107407c7f2f2b63/api/client/samples/crop_models
> 
> The brazil_soybeans folder is still there, with brazil_soybeans_model_overview.md
> 
> Want to remove that file and directory?

yup. 

> https://github.com/gro-intelligence/api-client/tree/e010553e35e37e330abbe8311107407c7f2f2b63/api/client/samples/crop_models
> 
> The brazil_soybeans folder is still there, with brazil_soybeans_model_overview.md
> 
> Want to remove that file and directory?

done

nemozen:
> "and" -> "or"

done

PR #122, opened 2019-11-03
Added endpoint that, given series query parameters, returns ranked so…

sempervirescent:
…urces.


api/client/lib.py
```diff
@@ -443,6 +443,24 @@ def get_data_series(access_token, api_host, **selection):
         raise Exception(resp.text)
 
 
+# TODO: Add code to handle no query parameters call to v2/available/sources. It is hanging at the moment.
+
+
+def get_source_ranking(access_token, api_host, series):
```
nemozen, line 455: this function returns the same thing as rank_series_by_source....

sempervirescent: No, it does not. get_source_ranking is called inside rank_series_by_source. The first one calls the endpoint directly and returns an array of ranked sources. The latter an array of series with all information. I separated them so we can also get the ranking only directly, without the rest of the data.

sempervirescent:
@itsjohnward: clarified all comments by @nemozen. If you have a chance to take a look, would you let me know if you have any concerns? If not, could we merge? Thank you.

api/client/__init__.py
```diff
@@ -40,9 +40,11 @@ def lookup_belongs(self, entity_type, entity_id):
         return lib.lookup_belongs(self.access_token, self.api_host,
                                   entity_type, entity_id)
 
+    def get_source_ranking(self, series):
```
nemozen: this method Client.get_source_ranking is not used ...

api/client/__init__.py
```diff
@@ -40,9 +40,11 @@ def lookup_belongs(self, entity_type, entity_id):
...
+        return lib.get_source_ranking(self.access_token, self.api_host, series)
+
     def rank_series_by_source(self, series_list):
-        return lib.rank_series_by_source(self.access_token, self.api_host,
-                                         series_list)
+        return lib.rank_series_by_source(self.access_token, self.api_host, series_list)
```
nemozen: lines should be < 80 chars as much as possible.

api/client/batch_client.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, api_host, access_token):
         super(BatchClient, self).__init__(api_host, access_token)
         self._logger = lib.get_default_logger()
         self._http_client = AsyncHTTPClient()
+        self.path = ''
```
nemozen: unused variable

api/client/batch_client.py
```diff
@@ -80,20 +81,33 @@ def get_data(self, url, headers, params=None):
 
     @gen.coroutine
     def get_data_points(self, **selection):
-        """Get all the data points for a given selection, which is some or all
-        of: item_id, metric_id, region_id, frequency_id, source_id,
-        partner_region_id. Additional arguments are allowed and ignored.
+        """Get all the data points for a given selection, which is some or all of: item_id, metric_id, region_id,
+        frequency_id, source_id, partner_region_id. Additional arguments are allowed and ignored.
```
nemozen: prefer to keep lines < 80  chars and also to avoid whitespace changes unless they are an improvement

api/client/batch_client.py
```diff
@@ -80,20 +81,33 @@ def get_data(self, url, headers, params=None):
...
+        headers = {'authorization': 'Bearer ' + self.access_token}
+        url = '/'.join(['https:', '', self.api_host, 'v2/available/sources'])
+        params = dict((k + 's', v) for k, v in iter(list(
+            lib.get_params_from_selection(**selection).items())))
+        resp = yield self.get_data(url, headers, params)
+        response = selection
+        response['ranked_sources'] = json_decode(resp)
+        raise gen.Return(response)
+
+    def batch_async_get_ranked_sources(self, batched_args, output_list=None, map_result=None):
```
nemozen: series ranking does not benefit from batch client's async processing... so we don't need a different implementation. let's just make BatchCLient a subclass of GroClient and it will inherit this functionality

api/client/lib.py
```diff
@@ -443,6 +443,24 @@ def get_data_series(access_token, api_host, **selection):
...
+    :return: List of sources that match the series parameters, sorted by rank.
+    """
+    url = '/'.join(['https:', '', api_host, 'v2/available/sources'])
+    headers = {'authorization': 'Bearer ' + access_token}
+    params = dict((k + 's', v) for k, v in iter(list(
+        get_params_from_selection(**series).items())))
+    return get_data(url, headers, params).json()
+
+
 def rank_series_by_source(access_token, api_host, series_list):
```
nemozen, line 474: added support for time ranges in  https://github.com/gro-intelligence/api-client/pull/125 
and also separating finding frmo ranking..

nemozen (changes requested):

api/client/__init__.py
sempervirescent: Reverted.

api/client/batch_client.py
sempervirescent: Reverted.

api/client/batch_client.py
sempervirescent: Removed.

api/client/__init__.py
sempervirescent: It is used in gro derived.

api/client/batch_client.py
sempervirescent: @nemozen we discussed this yesterday. I have both individual api calls tested in gro derived, as well as batch. Batch takes 28s as opposed to 1min 15sec on average for individual api calls. Are you suggesting that I should keep the individual api calls although they take a lot longer?

api/client/lib.py
sempervirescent: @nemozen I don't understand this comment.

api/client/batch_client.py
sempervirescent: I am not seeing any functionality in GroClient that deals with ranking. What did you refer to when you mentioned that we can inherit this functionality from there? Thx

api/client/lib.py
nemozen: just fyi  that this function was changing to also support time range as part of the ranking, which it didn't before.  merged into this now.

api/client/batch_client.py
sempervirescent: This was taken offline.

nemozen (approved):
made a few changes and retested, looks good. the batch and and non-batch version  return the same results.

```
for data_series in client.rank_series_by_source(series_list):
    print(data_series)

for batch_resp in client.batch_async_rank_series_by_source(
        [{'series_list': series_list}]):
    for data_series in batch_resp:
        print(data_series)

```

api/client/gro_client.py
```diff
@@ -129,6 +129,7 @@ def find_data_series(self, **kwargs):
             search_results.append(
                 self.search('regions', kwargs['partner_region'])[:cfg.MAX_RESULT_COMBINATION_DEPTH])
             keys.append('partner_region_id')
+        all_data_series = []
```
sempervirescent, line 132: @nemozen The changes in these files seem to come from another unrelated PR https://github.com/gro-intelligence/api-client/pull/127

api/client/lib.py
```diff
@@ -452,6 +452,25 @@ def get_data_series(access_token, api_host, **selection):
         raise Exception(resp.text)
 
 
+def get_source_ranking(access_token, api_host, series):
```
sempervirescent, line 455: @nemozen Based on our conversation on chat, do we want to add back the reference to this that was removed from Client?
```
def get_source_ranking(self, series):
    return lib.get_source_ranking(self.access_token, self.api_host, series)
```

nemozen:
yes it's a prereq

nemozen:
no i think find_data_series does what you want..

PR #123, opened 2019-11-04
Clews 22325

seungatGro:
add option to omit historical regions within lib.get_descendant_regions, off by default. 

blocked by pr https://bitbucket.org/grointelligence/gro/pull-requests/12725/clews-22325-include-historical-field-when/diff

see 3) in this issue https://grointelligence.atlassian.net/browse/CLEWS-22325

> get_decedents should have the option to exclude historical regions from the response. This should not be the default option, but it should be an option.

### DIFF
--- a/api/client/__init__.py
+++ b/api/client/__init__.py
@@ -50,6 +50,6 @@ class Client(object):
     def get_geojson(self, region_id):
         return lib.get_geojson(self.access_token, self.api_host, region_id)
 
-    def get_descendant_regions(self, region_id, descendant_level=None):
+    def get_descendant_regions(self, region_id, descendant_level=None, includeHistorical=True):
         return lib.get_descendant_regions(self.access_token, self.api_host,
-                                          region_id, descendant_level)
+                                          region_id, descendant_level, includeHistorical)
--- a/api/client/__init__.py
+++ b/api/client/__init__.py
@@ -50,6 +50,6 @@ class Client(object):
     def get_geojson(self, region_id):
         return lib.get_geojson(self.access_token, self.api_host, region_id)
 
-    def get_descendant_regions(self, region_id, descendant_level=None, includeHistorical=True):
+    def get_descendant_regions(self, region_id, descendant_level=None, include_historical=True):
         return lib.get_descendant_regions(self.access_token, self.api_host,
-                                          region_id, descendant_level, includeHistorical)
+                                          region_id, descendant_level, include_historical)
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -873,6 +873,8 @@ def get_descendant_regions(access_token, api_host, region_id,
     region_id : integer
     descendant_level : integer
         The region level of interest. See REGION_LEVELS constant.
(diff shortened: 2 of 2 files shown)


api/client/__init__.py
```diff
@@ -50,6 +50,6 @@ def get_geo_centre(self, region_id):
     def get_geojson(self, region_id):
         return lib.get_geojson(self.access_token, self.api_host, region_id)
 
-    def get_descendant_regions(self, region_id, descendant_level=None):
+    def get_descendant_regions(self, region_id, descendant_level=None, includeHistorical=True):
         return lib.get_descendant_regions(self.access_token, self.api_host,
-                                          region_id, descendant_level)
+                                          region_id, descendant_level, includeHistorical)
```
alexsherman: I think include_historical should be snake_case here, as well

api/client/lib.py
```diff
@@ -899,11 +899,11 @@ def get_descendant_regions(access_token, api_host, region_id,
     region = lookup(access_token, api_host, 'regions', region_id)
     for member_id in region['contains']:
         member = lookup(access_token, api_host, 'regions', member_id)
-        if descendant_level == member['level']:
+        if descendant_level == member['level'] and (include_historical or not member['historical']):
             descendants.append(member)
         elif member['level'] < descendant_level:
```
alexsherman, line 908: Minor - not sure we want to recursively drill down into historical regions' descendants if include_historical is false - I guess it depends on whether a historical region has non-historical descendants, and what that would mean. But presumably, even if a historical region has non-historical descendants, if those descendants are truly non-historical, than the non-historical ancestors should contain those, so they will be added to descendants anyways.

api/client/__init__.py
seungatGro: good catch, pushed change

api/client/lib.py
seungatGro: agree, should not drill down historical regions. pushed change

alexsherman (changes requested):
should update parameter list in comment for get_descendant_regions

seungatGro:
> should update parameter list in comment for get_descendant_regions

had missed this. done

alexsherman (approved):

api/client/lib.py
```diff
@@ -855,7 +855,7 @@ def get_geojson(access_token, api_host, region_id):
 
 
 def get_descendant_regions(access_token, api_host, region_id,
-                           descendant_level):
+                           descendant_level, include_historical):
```
itsjohnward: Need to include a default `include_historical=True` here as well to make it optional.

We encourage using the Client classes, in which case there's a default already, but some folks are using the lib directly and this constitutes a breaking change since it adds a new required param.

itsjohnward (changes requested):

api/client/lib.py
seungatGro: good catch! done

itsjohnward (approved):

nemozen (approved):

PR #124, opened 2019-11-06
CLEWS-22325: add console warning

seungatGro:
pending pr on api server: https://bitbucket.org/grointelligence/gro/pull-requests/12765/clews-22325-return-region-historical/diff

tested with local response:

`
  File "<stdin>", line 3, in <module>
Warning: Some of the regions in your data call are historical, with boundaries that may be outdated. The regions may have overlapping values with current regions
`


api/client/lib.py
```diff
@@ -438,7 +438,11 @@ def get_data_series(access_token, api_host, **selection):
     params = get_params_from_selection(**selection)
     resp = get_data(url, headers, params)
     try:
-        return resp.json()['data']
+        response = resp.json()['data']
+        if any(map(lambda series: 
+            series['historical_region'] or series['historical_partner_region'], response)):
+            raise Warning('Some of the regions in your data call are historical, with boundaries that may be outdated. The regions may have overlapping values with current regions')
```
itsjohnward: Don't need to raise an exception. Can just use `logger.warning()`

api/client/lib.py
```diff
@@ -438,7 +438,11 @@ def get_data_series(access_token, api_host, **selection):
     params = get_params_from_selection(**selection)
     resp = get_data(url, headers, params)
     try:
-        return resp.json()['data']
+        response = resp.json()['data']
+        if any(map(lambda series: 
+            series['historical_region'] or series['historical_partner_region'], response)):
```
itsjohnward: Use `series.get('historical_region', False) or series.get('historical_partner_region', False)` instead of square brackets so we can make this field optional.

itsjohnward (changes requested):

api/client/lib.py
```diff
@@ -438,7 +438,11 @@ def get_data_series(access_token, api_host, **selection):
     params = get_params_from_selection(**selection)
     resp = get_data(url, headers, params)
     try:
-        return resp.json()['data']
+        response = resp.json()['data']
+        if any(map(lambda series:
```
itsjohnward: List comprehension is preferred over maps: `any(series.get('historical_region', False) for series in response)`

itsjohnward (changes requested):

api/client/lib.py
seungatGro: done

api/client/lib.py
seungatGro: done

api/client/lib.py
seungatGro: done

itsjohnward: `if not logger` is failing on Shippable because logger's not defined.

Just remove the conditional and do `logger = get_default_logger()`

seungatGro: done

api/client/lib.py
itsjohnward: Should be False, not True

itsjohnward: Also the unit test here https://github.com/gro-intelligence/api-client/blob/fbba757170e47c3075490b8608e26ee3fffef2af/api/client/lib_test.py#L11 has these responses mocked out as strings instead of as dicts.

Need to update that to match the expected format

seungatGro: > Should be False, not True

edit: false call. default value = false. updated

itsjohnward (approved):

itsjohnward (approved):

PR #125, opened 2019-11-07
GroClient add_data_series() improvement

nemozen:
- take time ranges into account when ranking available series
- refactor add_data_series to use a separate find_data_series function

### DIFF
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -126,6 +126,7 @@ class GroClient(Client):
             search_results.append(
                 self.search('regions', kwargs['partner_region'])[:cfg.MAX_RESULT_COMBINATION_DEPTH])
             keys.append('partner_region_id')
+        all_data_series = []
         for comb in itertools.product(*search_results):
             entities = dict(list(zip(keys, [entity['id'] for entity in comb])))
             data_series_list = self.get_data_series(**entities)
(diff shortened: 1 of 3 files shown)


api/client/gro_client.py
```diff
@@ -126,9 +131,25 @@ def add_data_series(self, **kwargs):
             data_series_list = self.get_data_series(**entities)
             self._logger.debug("Found {} distinct data series for {}".format(
                 len(data_series_list), entities))
+            # temporal coverage affects ranking so add time range if specified.
+            for data_series in data_series_list:
+                if kwargs.get('start_date'):
```
itsjohnward, line 136: These dates are getting filtered out in `get_params_from_selection()`.

Also even if it wasn't, `dict((k + 's', v) for k, v in iter(list(get_params_from_selection(**series).items())))` would be converting the keys to `start_dates` and `end_dates` which wouldn't be valid inputs.

api/client/gro_client.py
```diff
@@ -126,9 +131,25 @@ def add_data_series(self, **kwargs):
...
+                    data_series['start_date'] = kwargs['start_date']
+                if kwargs.get('end_date'):
+                    data_series['end_date'] = kwargs['end_date']
             for data_series in self.rank_series_by_source(data_series_list):
```
itsjohnward, line 140: Do note rank_series_by_source() still doesn't work as expected when passing in a full data series. https://grointelligence.atlassian.net/browse/CLEWS-17127

Only accepts metric/item/region/frequency. No source_id.

itsjohnward (changes requested):

api/client/gro_client.py
nemozen: Right. Actuall it does accept (and correctly ignore) source_id but  the source_name remains in the output and mismatches. Fixed in lib.py

tolothy (approved):
tested with 
`groClient.add_data_series(**{
                'metric': 'Area planted (area)',
                'item': 'Corn',
                'region': 'United States', 
                'start_date': '2019-01-01',
                'end_date': '2019-12-31',
})`

Prior to the change `{u'metric_id': 2580001, u'source_name': u'WASDE', u'region_id': 1215, u'end_date': u'2020-08-31T00:00:00.000Z', u'item_name': u'Corn', u'partner_region_name': u'World', u'frequency_id': 9, 'source_id': 106, u'region_name': u'United States', u'partner_region_id': 0, u'item_id': 274, u'start_date': u'2008-09-01T00:00:00.000Z', u'metric_name': u'Area Planted (area)'}` is added.

After the change `{'partner_region_id': 0, 'source_id': 25, 'item_id': 274, 'metric_name': 'Area Planted (area)', 'metric_id': 2580001, 'frequency_id': 9, 'partner_region_name': 'World', 'start_date': '2019-01-01', 'region_id': 1215, 'item_name': 'Corn', 'end_date': '2019-12-31', 'region_name': 'United States'}` is added, which means added start_date and end_date kwargs were taken into consideration when data series were ranked.

PR #126, opened 2019-11-08
Crop model GDD

nemozen:
CropModel.growing_degree_days()


nemozen:
branched off https://github.com/gro-intelligence/api-client/pull/125

api/client/crop_model.py
```diff
@@ -69,3 +70,51 @@ def compute_crop_weighted_series(self,
...
+                            start_date, end_date):
+        """Get Growing Degree Days (GDD) for a region.
+
+        Growing degree days (GDD) are a weather-based indicator that
+        allows for assessing crop phenology and crop development,
+        based on heat accumulation. GDD for one day is defined as
+        [(T_max + T_min)/2 - T_base], and the GDD over a longer time
+        interval is the sum of the GDD over all days in the interval.
+
+        The region can be any region of the Gro regions, from a point
```
jeanzhaoyi, line 181: What if only Tmean is available for a region (say using LST) while daily Tmin and Tmax were not available or incomplete (ghcn stations)?

nemozen: Yes good question.  The idea is that we use find_data_series to get the best version of GDD, and that depends on the combination of search ranking and series availability.

If the region is specific and the daily max and min exists for that exact region,  e.g. a weather station  with GHCN data, then it does exactly what you would expect.   If not , then it depends on what is avialable.. If we don't have max and min temp but we do have the daily mean temperature, then  that should be the closest matching item and metric, and taking the mean of that should be ok. Or if we have more samples uniformly throughout the day  e.g. temperature every hour, then takign the average ofa all of them should be ok too. But note that *land surface* temperature is different than *air*  temperature so that would be wrong. 
Similarly, if the region is higher level, then it should give preference to weather stations in that district, rather than picking other lst at the district level.

Updataed the code to make the generalization clearer.

jeanzhaoyi: Often there are some GHCN stations with available data. The issue is that GHCN station data tend to be highly sparse temporally. For example, there might be data for 2 years or a few days a year, while the desired time period is spreading over 10 years. I think there needs to be a judgement call on a minimum threshold of temporal coverage for a region required to use the data from a source. Like you said, air temperature is very different from land surface temperature, so we need to be consistent with choosing one temperature source for a given spatial extent (likely a list of districts or provinces) and time period.

nemozen: Yup, I was planning on adding coverage threshold later. But I've added it to this PR.

api/client/crop_model.py
```diff
@@ -69,3 +70,75 @@ def compute_crop_weighted_series(self,
...
+        Parameters
+        ----------
+        region_name: string, required
+        base_temperature: number, required
+        start_date: '%Y-%m-%d' string, required
+        end_date: '%Y-%m-%d' string, required
+        min_temporal_coverage: float, optional
+
+        """
+        for tmax in self.find_data_series(
```
jeanzhaoyi: Since we are not specifying the source_id here, it could be LST or GHCN or some other future sources right?

nemozen: Right. It will use  the source ranking for that time range and region.

jeanzhaoyi: In that case, a user would bear the risk of having their daily mean temperature calculated from mixed sources. The risk being that certain regions or dates with more availability of one source have a more biased temperature than others, since LST tends to have higher Tmean than the GHCN Tmean. I think the current script is fine if a user can ignore such biases or uncertainties attributed to various data availability scenarios. Meanwhile, a user should be able to specify one particular source to use exclusively as well, like an argument `source_name` or `source_id` with default value Null, in which case the source ranking decides the source(s).

nemozen: Right.. 

growing_degree_days()  uses find_data_series()  entity search and series availability  to  avoid the user having to know which specific series they need, which depends on the region and the date ranges. 

I've separated the lower level calculation into a separate function compute_gdd() so a user who already knows how to select specific series can do it directly.

api/client/crop_model.py
```diff
@@ -69,3 +119,83 @@ def compute_crop_weighted_series(self,
...
+                           ['region_id', 'metric_id', 'frequency_id',
+                            'start_date', 'end_date']).mean()
+        duration = datetime.strptime(end_date, '%Y-%m-%d') - \
+                   datetime.strptime(start_date, '%Y-%m-%d')
+        coverage_threshold = min_temporal_coverage * duration.days
+        if tmean.value.size < coverage_threshold:
+            raise Exception("Insufficient temporal coverage for GDD, " + \
+                            "{} < {} data points available".format(
+                                tmean.value.size, coverage_threshold))
+        gdd_values = tmean.value - base_temperature
```
hongwyx: A few comments to ensure that the function works according to context:

The function currently allows for GDD to be unbounded; however, according to the source below, GDD shouldn't be negative, and should have an upper bound. At the very least I think we should not let it be negative, by modifying 

`gdd_values = tmean.value - base_temperature`

to 

`gdd_values = tmean.value.apply(lambda x: min(x - base_temperature, 0))`

and also require base_temperature to be non-negative (or within a certain range), otherwise raise exception.

When there is no temperature data for a particular day, it would be better to raise a warning so that the user knows that the GDD is not accumulating due to missing data, instead of ineligible temperature range.

Also, not sure if all temperature data series that we pull share the same unit (Fahrenheit/Celsius?) We should specify that in case the user is confused about what to use as the base temperature. (Do we have a unit converter function?)

(Additionally, we could also require start and end date to be within the same growing season.)

Attached is my script where I tested a few input combinations. I modified the import path/generator syntax (`next(...)` instead of `(...).next()` )to make it work with Python 3.


Sources:
http://agron-www.agron.iastate.edu/courses/Agron541/classes/541/lesson02b/2b.1.1.html

jeanzhaoyi: If the temperature at a given day, say `tmean.value`, is lower than the `base_temperature`, the contribution to GDD on that day should be zero.
Here's how I calculated GDD: 
```
(( (temp % up_thresh) * (temp < up_thresh) + 
(temp > up_thresh) * up_thresh) - lower_thresh) * (temp >= lower_thresh)
```

hongwyx: Modified to match your expression.

nemozen: thanks!

- changed each gdd to >= 0 
- added support for upper_temperature_cap
- changed default min_temporal_coverage to 1.0 so by default we do not allow any missing days at all
- added warning when time range > 1 year

jeanzhaoyi (approved):
Looks good to me =)

PR #127, opened 2019-11-12
in find_data_series(), yield all possible series  in order instead of…

nemozen:
… just the highest ranked one

add_data_series() will still only use the first one, but this gives
more flexibility when using the lower level function find_data_series


nemozen:
merged into https://github.com/gro-intelligence/api-client/pull/122/commits

PR #128, opened 2019-11-14
V1.40.3 branch

itsjohnward:
The `get_logger()` feature added in https://github.com/gro-intelligence/api-client/pull/111/files#diff-603e79178b8e522d0e366a9fca477b24R44 needs to be included in a released version of the client library so prevent plant can utilize it. Tagging a new version so the update will be received.

### DIFF
--- a/api/client/samples/prevented_plant/README.md
+++ b/api/client/samples/prevented_plant/README.md
@@ -1,6 +1,6 @@
 # Gro API Client example: Prevented Plant
 
-1. Follow the [instructions to install the Gro API Client](../../../../README.md)
+1. Follow the [instructions to install the Gro API Client](../../../../README.md). NOTE: requires a minimum client version of v1.40.3.
 2. Download the [requirements file in this directory](./requirements.txt)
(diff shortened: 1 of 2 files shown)


JGrinst (approved):
Good catch!

ash-gro (approved):
Looks good to me. In retrospect, get_logger() is a new feature of the main client, so should probably not have added as part of this branch which deals with rewriting one of usage examples (not sure Nemo would agree though).

itsjohnward:
> Looks good to me. In retrospect, get_logger() is a new feature of the main client, so should probably not have added as part of this branch which deals with rewriting one of usage examples (not sure Nemo would agree though).

I agree - I think it's a good approach to have a feature branch merged and tagged prior to merging the updated sample. Make it a pre-req if needing to use a feature from it

nemozen:
> > Looks good to me. In retrospect, get_logger() is a new feature of the main client, so should probably not have added as part of this branch which deals with rewriting one of usage examples (not sure Nemo would agree though).
> 
> I agree - I think it's a good approach to have a feature branch merged and tagged prior to merging the updated sample. Make it a pre-req if needing to use a feature from it

agreed.  this was a backward compatible change and the assumption is that people running new version of sample code would have the most recent client release

PR #129, opened 2019-11-15
get_descendant_regions: absence of descendant_level is equivalent to "any"

nemozen:
this allows us to get all the levels

### DIFF
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -919,10 +919,10 @@ def get_descendant_regions(access_token, api_host, region_id,
     for member_id in region['contains']:
         member = lookup(access_token, api_host, 'regions', member_id)
         if (not include_historical and member['historical']):
-            pass
-        elif descendant_level == member['level']:
+            continue
+        if not descendant_level or descendant_level == member['level']:
             descendants.append(member)
-        elif member['level'] < descendant_level:
+        if not descendant_level or member['level'] < descendant_level:
             descendants += get_descendant_regions(
                 access_token, api_host, member_id, descendant_level, include_historical)
     return descendants
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -872,7 +872,7 @@ def get_geojson(access_token, api_host, region_id):
 
 
 def get_descendant_regions(access_token, api_host, region_id,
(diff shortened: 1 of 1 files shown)


api/client/lib.py
```diff
@@ -919,10 +919,10 @@ def get_descendant_regions(access_token, api_host, region_id,
     for member_id in region['contains']:
         member = lookup(access_token, api_host, 'regions', member_id)
         if (not include_historical and member['historical']):
-            pass
-        elif descendant_level == member['level']:
+            continue
+        if not descendant_level or descendant_level == member['level']:
```
itsjohnward, line 923: `descendant_level` is still a required positional param, so you'd need to explicitly set it to `None` or `False` or `0`.

Could give it a default value of `None` so we could do something like `get_descendant_regions(1215)`. Or is that not desired since it would be making the more expensive case the default?

itsjohnward (approved):

api/client/lib.py
nemozen: Added default. Note it was already there in the calling Client method: https://github.com/gro-intelligence/api-client/blob/7dc79b81e9d9311b2ec4ad17ea53b126c6672ea7/api/client/__init__.py#L53

itsjohnward: Also if you set a default in the lib, need to also set a default in Client, or it'll still throw the required positional argument error if trying to use it without descendant_level

api/client/lib.py
```diff
@@ -919,10 +919,10 @@ def get_descendant_regions(access_token, api_host, region_id,
     for member_id in region['contains']:
         member = lookup(access_token, api_host, 'regions', member_id)
         if (not include_historical and member['historical']):
-            pass
-        elif descendant_level == member['level']:
```
seungatGro, line 923: ah sorry wrong assumption that descendant_level was required

seungatGro (approved):

api/client/lib.py
nemozen: note that this is the more general low level function where you know your region ids and region level ids. 

for simplicity, users can use wrappers like get_provinces(<country_name>)

PR #135, opened 2019-12-02
GDH

nemozen:


stephenbernardi (commented):
I think it would be most useful to have the output be a dataframe, and all the entity names located in the column header of each series

nemozen:
> I think it would be most useful to have the output be a dataframe, and all the entity names located in the column header of each series

done

api/client/gro_client.py
```diff
@@ -82,6 +88,48 @@ def get_df(self):
                 self._data_frame = self._data_frame.merge(tmp, how='outer')
         return self._data_frame
 
+    def GDH(self, gdh_selection):
+        """Wrapper for get_data_points with alternative input and output style.
+
+        The selection of data series to retrieve is encoded in a
+        'gdh_seletion' string of the form
+        <metric_id>-<item_id>-<region_id>-<source_id>-<frequency_id>
```
JGrinst: This is mentioned in the PRD comments, but for consistency we should probably also always include partner_region in the string even if a partner_region hasn't be explicitly selected.

JGrinst: @nemozen also units.

In @Fryolator2 's local example, the string requires <metric_id>-<item_id>-<region_id>-<source_id>-<frequency_id> and then has the options for partner region and units on the end. If these are not specified, they are assumed to be 'world' and the default.

This same functionality, or to always require these two would be important.

api/client/gro_client.py
```diff
@@ -82,6 +88,54 @@ def get_df(self):
...
+        gdh_selection: string
+
+        Returns:
+        ------
+        DataFrame
+
+            subset of the main DataFrame get_df() with only the requested series.
+
+        """
+        entity_keys = ['metric_id', 'item_id', 'region_id', 'partner_region_id',
```
muzigao: @JGrinst  @nemozen 
I added partner_region as required entity key , and included unit in output 

I'm think it could be a separate improvement to support user requested params like custom unit, revisions etc

api/client/gro_client.py
```diff
@@ -82,6 +88,54 @@ def get_df(self):
...
+        pairs.append(("unit", unit_name))
+        cols = pandas.MultiIndex.from_tuples([tuple(pair[1] for pair in pairs)],
+                                             names=tuple(pair[0] for pair in pairs))                                                          
+        filters = ((df['item_id'] == selection['item_id']) & 
+                   (df['metric_id'] == selection['metric_id']) &
+                   (df['region_id'] == selection['region_id']) &
+                   (df['frequency_id'] == selection['frequency_id']))
+        if 'partner_region_id' in df.columns:
+            filters = (filters & (df['partner_region_id'] == selection['partner_region_id']))
+        series = pandas.DataFrame(
```
muzigao: @nemozen  What's the df index issue you mentioned earlier? 

but right now output will break if data is from expansion , this filter will exclude all rows unless it's at_level, unless we aggregate output here.  not sure what's the desired behavior here.

api/client/gro_client.py
```diff
@@ -82,6 +88,37 @@ def get_df(self):
...
+            the main DataFrame get_df() with the get_data_points() results for requested series.
+
+        """
+        entity_keys = ['metric_id', 'item_id', 'region_id', 'partner_region_id',
+                       'source_id', 'frequency_id']
+        entity_ids = [int(x) for x in gdh_selection.split('-')]
+        selection = dict(zip(entity_keys, entity_ids))
+        self.add_single_data_series(selection)
+        df = self.get_df()
+        return df
```
muzigao: @nemozen  output format is removed, returning raw df now

stephenbernardi (approved):
tested the example locally and it looks good. I assume we are addressing the output format separately?

api/client/gro_client.py
```diff
@@ -224,6 +230,38 @@ def get_data_points(self, **selections):
         # Return data points in input units if not unit is specified
         return data_points
 
+    def GDH(self, gdh_selection):
```
muzigao: @nemozen @JGrinst   https://github.com/gro-intelligence/api-client/pull/168 added support to optional params e.g unit_id, show_revisions etc

nemozen:
> tested the example locally and it looks good. I assume we are addressing the output format separately?

yes

api/client/gro_client.py
nemozen: looks good merged

PR #136, opened 2019-12-05
simplify rank series by source

nemozen:
we remove teh bit that deals with duplicates in the list of series. this is over-optimization.
the caller can easily avoid duplicates, and if not, it's relatively harmless in ranking

### DIFF
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -336,6 +336,9 @@ def get_source_ranking(access_token, api_host, series):
 def rank_series_by_source(access_token, api_host, series_list):
     for series in series_list:
         try:
+            # Remove source if selected, to consider all sources.
+            series.pop('source_name', None)
+            series.pop('source_id', None)
             source_ids = get_source_ranking(access_token, api_host, series)
         except ValueError:
             continue  # empty response
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -334,15 +334,7 @@ def get_source_ranking(access_token, api_host, series):
 
 
 def rank_series_by_source(access_token, api_host, series_list):
-    # We sort the internal tuple representations of the dictionaries because
-    # otherwise when we call set() we end up with duplicates if iteritems()
-    # returns a different order for the same dictionary. See test case.
-    selections_sorted = set(tuple(sorted(
-        [k_v for k_v in iter(list(single_series.items()))
-         if k_v[0] not in ('source_id', 'source_name')],
-        key=lambda x: x[0])) for single_series in series_list)
-
-    for series in map(dict, selections_sorted):
+    for series in series_list:
         try:
             source_ids = get_source_ranking(access_token, api_host, series)
         except ValueError:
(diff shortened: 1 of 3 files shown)


muzigao (approved):

api/client/lib.py
```diff
@@ -334,15 +334,7 @@ def make_key(key):
...
-    # We sort the internal tuple representations of the dictionaries because
-    # otherwise when we call set() we end up with duplicates if iteritems()
-    # returns a different order for the same dictionary. See test case.
-    selections_sorted = set(tuple(sorted(
-        [k_v for k_v in iter(list(single_series.items()))
-         if k_v[0] not in ('source_id', 'source_name')],
-        key=lambda x: x[0])) for single_series in series_list)
-
-    for series in map(dict, selections_sorted):
+    for series in series_list:
```
muzigao, line 337: need to update test_rank_series_by_source

api/client/lib.py
```diff
@@ -334,15 +334,7 @@ def make_key(key):
 
 
 def rank_series_by_source(access_token, api_host, series_list):
-    # We sort the internal tuple representations of the dictionaries because
-    # otherwise when we call set() we end up with duplicates if iteritems()
-    # returns a different order for the same dictionary. See test case.
-    selections_sorted = set(tuple(sorted(
-        [k_v for k_v in iter(list(single_series.items()))
-         if k_v[0] not in ('source_id', 'source_name')],
```
itsjohnward, line 342: Now we've got that source_id-source_name not matching behavior again. Need to filter out source_name, or else if you pass in a series object like `{..., 'source_id': 2, 'source_name': 'FAO'}` from `get_data_series()` it'll return something like

```py
[ {..., 'source_id': 2, 'source_name': 'FAO'},
  {..., 'source_id': 14, 'source_name': 'FAO'},
  {..., 'source_id': 26, 'source_name': 'FAO'} ]
```

itsjohnward: ^ Good thing to add a unit test for as well

itsjohnward (changes requested):

api/client/lib.py
itsjohnward: ref https://github.com/gro-intelligence/api-client/pull/125/commits/04a449523641783440035464c064034270ba9572

nemozen: Done. Cleaned up so pre-selected source is ignored since the goal is to consider all possible sources

itsjohnward (approved):
Tested with

```py
import os

for ranking in client.rank_series_by_source([{'metric_id': 860032, 'item_id': 274, 'region_id': 1215}]):
  print('ranking 1', ranking)

for ranking in client.rank_series_by_source([{'metric_id': 860032, 'item_id': 274, 'region_id': 1215, 'source_id': 2}]):
  print('ranking with source', ranking)

for ranking in client.rank_series_by_source([{'metric_id': 860032, 'item_id': 274, 'region_id': 1215, 'start_date': '2012-01-01'}]):
  print('ranking with date', ranking)

for data_series in client.find_data_series(item='corn', region='united states'):
  print('data_series', data_series)
```

Seeing source_name is excluded now and it works with historical regions metadata. LGTM

PR #140, opened 2019-12-09
add find_data_series and crop modeling methods

nemozen:


JGrinst (approved):
I think these look good to me. I think it would be much better for find_data_series to include an example input. I don't think it's very clear how to construct a query with just what's presented here.

@itsjohnward I think you said that's possible, but might take a bit of extra work to make sure the example isn't treated as a doctest

docs/api.rst
```diff
@@ -5,6 +5,8 @@ API Reference
 Basic Exploration
 *****************
 
+.. automethod:: api.client.gro_client.GroClient.find_data_series
```
itsjohnward: These appear in whatever order you put them in in this file, and similar functions are grouped together, in a progression, like: `lookup()`, `search()`, `search_and_lookup()`

Following that that same idea, `find_data_series()` should come after `get_data_series()`, since it's a combo of `search()` and `get_data_series()`

itsjohnward: The url in the "See also" section of the docstring isn't formatting nicely. Might need to put it in angle brackets? `<https://github.com/gro-intelligence/api-client/wiki/Data-Series-Definition>`? Can play around with that

docs/api.rst
```diff
@@ -50,3 +52,11 @@ Pandas Utils
 .. automethod:: api.client.gro_client.GroClient.add_single_data_series
 
 .. automethod:: api.client.gro_client.GroClient.get_data_series_list
+
+*************
+Crop Modeling
+*************
+
+.. automethod:: api.client.crop_model.CropModel.compute_weights
```
itsjohnward, line 60: These docstrings should just have single backticks, instead of triple

itsjohnward (changes requested):

docs/api.rst
JGrinst: I've updated the url to https://developers.gro-intelligence.com/data-series-definition.html in my fix-docs-links PR

itsjohnward: But it's not in this webpage in your PR since it's being added here. Didn't realize it renders weird when it's in the "See also" section (see https://developers.gro-intelligence.com/nemozen-patch-1/api.html ). Looks fine elsewhere in https://developers.gro-intelligence.com/fix-docs-links/api.html because we didn't have any instances of it being in "See also"

itsjohnward: Testing here to see if the angle brackets fix it: https://github.com/gro-intelligence/api-client/pull/140/commits/8080f1a9bac15845e3e6d6748cb9c82105cda8ef

itsjohnward: https://github.com/gro-intelligence/api-client/pull/140/commits/8a71aec167dc4577f7ab3dff3833a53afc0a4d5a this works to make it a link, but it messes up the `get_data_series()` format. Will just move the link up to the main description instead.

itsjohnward: https://github.com/gro-intelligence/api-client/pull/140/commits/3099c393d67e8c2870d1086606673112756b5abd done. This looks better.

api/client/gro_client.py
```diff
@@ -181,7 +181,22 @@ def add_single_data_series(self, data_series):
...
 
     def find_data_series(self, **kwargs):
-        """Attempts to find a matching data series for a combination of entities by name.
+        """Find the best possible  data series matching a combination of entities specified by name.
+
+        Example::
+
+            client.find_data_series(item="Corn",
+                                    metric="Futures Open Interest",
+                                    region="United States of America").next()
```
itsjohnward: `iter.next()` is Python 2

Update to `next(iter)`?

itsjohnward (changes requested):

itsjohnward:
> I think these look good to me. I think it would be much better for find_data_series to include an example input. I don't think it's very clear how to construct a query with just what's presented here.
> 
> @itsjohnward I think you said that's possible, but might take a bit of extra work to make sure the example isn't treated as a doctest

The way he did it here https://github.com/gro-intelligence/api-client/pull/140/commits/60dc0e9bdeae4ad19d6b4488f5182479a2f55364#diff-603e79178b8e522d0e366a9fca477b24R192 works. Would be picked up as a doctest if it were formatted like

```
>>> function(input)
output
```

But this isn't recognized as a valid doctest:

```
Example::
  function(input)

Outputs::
  output
```

docs/api.rst
nemozen: yup I saw they appear in the same order.   I like the thematic groups. within each group, I think it makes sense to go from simplest to hardest which is why I put this first.

api/client/gro_client.py

docs/api.rst
nemozen: done

api/client/gro_client.py
```diff
@@ -217,6 +217,7 @@ def find_data_series(self, **kwargs):
         See also
         --------
         get_data_series()
+        https://developers.gro-intelligence.com/data-series-definition.html
```
itsjohnward: I'd removed this here https://github.com/gro-intelligence/api-client/commit/3099c393d67e8c2870d1086606673112756b5abd because the formatting for the link doesn't work in the "See also" section. See the line-break here:
Screen Shot 2019-12-09 at 8 58 26 PM

itsjohnward (changes requested):

docs/api.rst
itsjohnward: I think there's also a flow from "entity" into "data series" though, which this breaks, since this sandwiches entity methods in-between the two data series methods

nemozen: hmm I was thinking of the flow being more like increasing complexity for the user, i.e. "I'm feeling lucky" is the first thing. THen when you get more sophisticated you start  looking at  lower level details. But that doesn't quite fit with lookup being before search, which is more like introducing more complicated concepts.  Ditto for data retrieval and pandas. So we can do taht in a separate pr

api/client/gro_client.py
itsjohnward: ^ This was still there, so I just pushed https://github.com/gro-intelligence/api-client/pull/140/commits/82b83d8a600578a0842088ab1875ff8784efe698 to remove

itsjohnward (approved):

PR #141, opened 2019-12-09
Switch to list_of_series and add formatting function back to single_series

itsjohnward:
See the test case: https://github.com/gro-intelligence/api-client/compare/CLEWS-20278-mvp?expand=1#diff-7ed38827a2c4e58f0dd776f8639dc95cR400

"list_of_series" output is in the format:

```py
{
  'series': { 'metricId': 1, 'itemId': 2, 'regionId': 3, 'belongsTo': { 'itemId': 22 } },
  'data': [
    ['2001-01-01', '2001-12-31', 123]
  ]
}
```

The formatter function translates that back into the normal "single_series" format API client's always used, just with the addition of the belongs_to property:

```py
[ {
    'start_date': '2001-01-01',
    'end_date': '2001-12-31',
    'value': 123,
    'unit_id': None,
    'reporting_date': None,
    'metric_id': 1,
    'item_id': 2,
    'region_id': 3,
    'partner_region_id': 0,
    'frequency_id': None,
    'source_id': None,
    'belongs_to': { 'item_id': 22 }
} ]
```

The list_of_series format reduces the amount of duplicate info transmitted in data requests, and the addition of the belongs_to property allows us to request multiple series in a single request while still being able to figure out what data came from which requested series in the event of expansions. For example:

```py
import os

request_response = { 100000021: set(), 100000022: set() }

  metric_id=30032,
  item_id=223,
  region_id=[100000021, 100000022],
  source_id=59,
  frequency_id=6,
  start_date='2010-01-01',
  end_date='2010-12-31'
):
  request_response[point['belongs_to']['region_id']].add(point['region_id'])

print(request_response)
```

outputs

```py
{100000021: {1096, 1105, 1075, 1011, 1205}, 100000022: {1092, 1032, 1162, 1134, 1167, 1210, 1051, 1183}}
```

If we didn't have the belongs_to property, we'd have points with region_ids 1096, 1105, 1075, 1011, 1205, 1092, 1032, 1162, 1134, 1167, 1210, 1051, and 1183, none of which are the regions we requested. We wouldn't know which ones belong to 100000021 and which ones belong to 100000022 without making additional queries.

## Further work:

I separated this PR out from my larger branch https://github.com/gro-intelligence/api-client/pull/116/files to serve as an MVP to get list_of_series out there and usable. Nice-to-haves in the other branch include:

1. If you request too many entities in a single request, automatically split it up into multiple requests instead of returning an error. In this PR, selecting an appropriate size request is the responsibility of the user
2. Group BatchClient requests together automatically under the hood. In this PR, you're able to make grouped requests using BatchClient in the same way you do with GroClient: making `region_id` or `item_id` into a list of integers instead of an integer. But if you pass in integers, it's an optimization we can make to group them for you without you doing anything.

### DIFF
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -115,12 +115,12 @@ class GroClient(Client):
                 }
             }]
 
-        Note: you can pass the output of :meth:`GroClient.get_data_series()` into
-        :meth:`GroClient.get_data_points()` to check what series exist for some selections and then
+        Note: you can pass the output of :meth:`api.client.gro_client.GroClient..get_data_series()` into
+        :meth:`api.client.gro_client.GroClient..get_data_points()` to check what series exist for some selections and then
         retrieve the data points for those series. See /api/client/samples/quickstart.py for an
         example of this.
 
-        :meth:`GroClient.get_data_points()` also allows passing a list of ids for metric_id,
+        :meth:`api.client.gro_client.GroClient..get_data_points()` also allows passing a list of ids for metric_id,
         item_id, and/or region_id to get multiple series in a single request. This can be faster if
         requesting many series.
 
@@ -209,9 +209,9 @@ class GroClient(Client):
 
     def get_data_series_list(self):
         """Inspect the current list of saved data series contained in the GroClient.
-        
-        For use with :meth:`GroClient.get_df()`. Add new data series to the list using
-        :meth:`GroClient.add_data_series()` and :meth:`GroClient.add_single_data_series()`.
+
+        For use with :meth:`api.client.gro_client.GroClient..get_df()`. Add new data series to the list using
+        :meth:`api.client.gro_client.GroClient..add_data_series()` and :meth:`api.client.gro_client.GroClient..add_single_data_series()`.
 
         Returns
         -------
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -455,7 +455,6 @@ class GroClient(Client):
         self._logger.info("Outputing to file: {}".format(filename))
         writer = unicodecsv.writer(open(filename, 'wb'))
         for point in self.get_data_points(**data_series):
-            print(point)
             writer.writerow([point['start_date'],
                              point['end_date'],
                              point['value'],
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -115,8 +115,8 @@ class GroClient(Client):
                 }
             }]
 
-        Note: you can pass the output of :meth:`get_data_series` into
-        :meth:`get_data_points` to check what series exist for some selections and then
+        Note: you can pass the output of :meth:`self.get_data_series` into
+        :meth:`GroClient.get_data_points` to check what series exist for some selections and then
         retrieve the data points for those series. See /api/client/samples/quickstart.py for an
         example of this.
 
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -47,17 +47,18 @@ class GroClient(Client):
 
 
     def get_df(self):
-        """Call get_data_points() for each saved data series and return as a combined dataframe.
+        """Call :meth:`~.get_data_points` for each saved data series and return as a combined
+        dataframe.
         
-        Note you must have first called either add_data_series() or add_single_data_series() to save
-        data series into the GroClient's data_series_list. You can inspect the client's saved list
-        using get_data_series_list().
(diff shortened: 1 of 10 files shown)


itsjohnward:
TODO: fix unit test

api/client/lib.py
```diff
@@ -349,12 +394,72 @@ def rank_series_by_source(access_token, api_host, series_list):
...
+    ...     'region_id': 3,
+    ...     'partner_region_id': 0,
+    ...     'frequency_id': None,
+    ...     'source_id': None,
+    ...     'belongs_to': { 'item_id': 22 }
+    ... } ]
+    True
+
+    """
+    if not isinstance(series_list, list):
```
seungatGro, line 453: should we handle errors in callers? `if response status code == 200 format_list_of_series(resp.data)`

seungatGro (commented):
implementation lgtm. think we can handle errors in callers of formatter, or change the formatter to a response handler

itsjohnward:
> TODO: fix unit test

Done

api/client/lib.py
itsjohnward: There's already all the error handling we want in `get_data()`. There we're handling timeouts, rate limits, migrations, etc. By the time it gets to this formatter function, if there's an error that's been raised by `get_data()` I just want to propagate it to the user.

api/client/lib.py
```diff
@@ -349,12 +394,77 @@ def rank_series_by_source(access_token, api_host, series_list):
...
+        } for point in series.get('data', [])]
+    return output
+
+
 def get_data_points(access_token, api_host, **selection):
+    assert 'metric_id' in selection, 'metric_id is required'
+    assert 'item_id' in selection, 'item_id is required'
+    assert 'region_id' in selection, 'region_id is required'
+    assert 'source_id' in selection, 'source_id is required'
+    assert 'frequency_id' in selection, 'frequency_id is required'
```
nemozen: This is not-backward compatible. currently get_data_points allows some of these to be unspecified which can be useful for advanced users.  Let's discuss offline.

api/client/lib.py
```diff
@@ -349,12 +394,77 @@ def rank_series_by_source(access_token, api_host, series_list):
             yield series_with_source
 
 
+def format_list_of_series(series_list):
```
nemozen: function name sounds like it goes the other way.. how about list_of_series_to_single_series

itsjohnward: Done

api/client/lib.py
itsjohnward: Removed for now. Not strictly necessary for this feature

nemozen (approved):

api/client/gro_client.py
```diff
@@ -387,18 +448,18 @@ def pick_random_data_series(self, selected_entities):
...
         """Output a data series to a CSV file."""
         self._logger.info("Using data series: {}".format(str(data_series)))
         self._logger.info("Outputing to file: {}".format(filename))
         writer = unicodecsv.writer(open(filename, 'wb'))
         for point in self.get_data_points(**data_series):
-            writer.writerow([point['start_date'], point['end_date'],
-                             point['value'] * point['input_unit_scale'],
-                             self.lookup_unit_abbreviation(point['input_unit_id'])])
-
+            print(point)
```
seungatGro: remove log

api/client/gro_client.py
```diff
@@ -387,18 +448,18 @@ def pick_random_data_series(self, selected_entities):
...
-
     # TODO: rename function to "write_..." rather than "print_..."
     def print_one_data_series(self, data_series, filename):
         """Output a data series to a CSV file."""
         self._logger.info("Using data series: {}".format(str(data_series)))
         self._logger.info("Outputing to file: {}".format(filename))
         writer = unicodecsv.writer(open(filename, 'wb'))
         for point in self.get_data_points(**data_series):
-            writer.writerow([point['start_date'], point['end_date'],
-                             point['value'] * point['input_unit_scale'],
```
seungatGro, line 399: removing the input unit scale multiplier?

seungatGro (approved):

api/client/gro_client.py
seungatGro: explained the transition from input unit id + scale -> unit id. left for backwards compatibility in lib, unnecessary elsewhere (just use value + unit id)

api/client/gro_client.py
itsjohnward: done

PR #142, opened 2019-12-10
Fix code blocks

JGrinst:
Documentation had bad formatting, so copied code would break

### DIFF
--- a/docs/quickstart-examples.md
+++ b/docs/quickstart-examples.md
@@ -37,10 +37,10 @@ If you have a Google account, or if you already have [Jupyter Notebook](https://
 
     ```py
     from api.client.gro_client import GroClient
-    client = GroClient('api.gro-intelligence.com', <YOUR_GROAPI_ACCESS_TOKEN>)
+    client = GroClient('api.gro-intelligence.com', <YOUR_TOKEN>)
     ```
 
-4. Replace the text YOUR_TOKEN with the token you copied from step one. Then click the "Run Cell" button to the left of the cell.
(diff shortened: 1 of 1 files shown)


docs/quickstart-examples.md
```diff
@@ -39,7 +39,8 @@ If you have a Google account, or if you already have [Jupyter Notebook](https://
     from api.client.gro_client import GroClient
     API_HOST = 'api.gro-intelligence.com'
     ACCESS_TOKEN = 'YOUR_TOKEN'
```
nemozen: 'YOUR_TOKEN' should <YOUR_TOKEN> instead.   Convention is that putting in quotes means it's a literal string. Angle brackets mean that is a description thereof rather than literally that.

nemozen: also these intermediate variables are not helping here... it's good practice if they are reused in many places, but here it's unnecessary clutter

nemozen (approved):

docs/quickstart-examples.md
JGrinst: > 'YOUR_TOKEN' should <YOUR_TOKEN> instead. Convention is that putting in quotes means it's a literal string. Angle brackets mean that is a description thereof rather than literally that.

I've faced failures, and seen others face failures because they didn't know they had to have these quotes around the key. So the quotes made it easier to just replace YOUR_TOKEN with your token and succeed. The new text does say that quotes are needed, but people also don't like to read, and the token is a string, no?

nemozen: Agreed. Belt and suspenders ;)

PR #144, opened 2019-12-12
 empty dataframe when no data is available

nemozen:
CLEWS-18123


stephenbernardi (approved):
looks good. tested with a nonexistent data series and get_df returns an empty dataframe.

PR #145, opened 2019-12-17
update docs with  new CropModel feature GDD

nemozen:


JGrinst (approved):

api/client/crop_model.py
```diff
@@ -11,16 +11,16 @@ class CropModel(GroClient):
 
     def compute_weights(self, crop_name, metric_name, regions):
         """Compute a vector of 'weights' that can be used for crop-weighted
-        average across regions.
+        average across regions, as in :meth:`~.compute_crop_weighted_series`.
```
itsjohnward, line 14: I just added these `:meth:~` links like in https://github.com/gro-intelligence/api-client/pull/141/files#diff-9adf8c29ac9359811cbf476c22bab58bR169 which make them clickable in the docs

api/client/crop_model.py
```diff
@@ -186,16 +219,23 @@ def growing_degree_days(self, region_name, base_temperature,
...
-        min_temporal_coverage: float, optional
-        upper_temperature_cap: float, optional
+        region_name : string
+        base_temperature : number
+        start_date : string
+            YYYY-MM-DD date
+        end_date : string
+            YYYY-MM-DD date
+        min_temporal_coverage : float, optional
+        upper_temperature_cap : number, optional
```
itsjohnward, line 233: I changed this from float to number. Doesn't have to be a float, right? Figured it should be the same type as base_temperature

itsjohnward (approved):

PR #150, opened 2019-12-24
add option show_revisions to get_df

eliaschneider:
Add boolean variable in get_df to add show_revision=True in dictionary data_series. 
using this option we can retrieve data with same start_date but different reporting_date.

### DIFF
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -62,7 +62,7 @@ class GroClient(Client):
         """
         while self._data_series_queue:
             data_series = self._data_series_queue.pop()
-	    if show_revisions: 
+            if show_revisions: 
                data_series['show_revisions'] = True
             tmp = pandas.DataFrame(data=self.get_data_points(**data_series))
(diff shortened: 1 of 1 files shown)


nemozen (approved):
LGTM
Tested it with futures_contracts which has 3 different contracts. As expected:
with revisions, get_df returns a frame with  344881 rows
without revisions it has 785 rows

api/client/gro_client.py
```diff
@@ -62,6 +62,8 @@ def get_df(self):
         """
         while self._data_series_queue:
             data_series = self._data_series_queue.pop()
```
tolothy, line 64: do we want to cover the cases that the original input entity has `show_revisions=True`?

nemozen: yes for the user who wants ot directly control the details, you can put show_revisions on individual data series  via add_single_data_series directly. but this at get_df level, gives a shortcut to just do it on all series.

PR #151, opened 2019-12-24
Batch client get df

nemozen:
Creates the same data frame as the non-async GroClient but using async data retrieval for better performance.

### DIFF
--- a/api/client/batch_client.py
+++ b/api/client/batch_client.py
@@ -79,14 +79,6 @@ class BatchClient(GroClient):
                         url, retry_count, e.response.error))
 
     @gen.coroutine
-    def get_data_points_generator(self, **selection):
-        headers = {'authorization': 'Bearer ' + self.access_token}
-        url = '/'.join(['https:', '', self.api_host, 'v2/data'])
-        params = lib.get_data_call_params(**selection)
-        resp = yield self.get_data(url, headers, params)
-        raise gen.Return(json_decode(resp))
-
-    @gen.coroutine
     def get_data_points(self, **selection):
         """Get all the data points for a given selection, which is some or all
         of: item_id, metric_id, region_id, frequency_id, source_id,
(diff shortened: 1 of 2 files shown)


api/client/batch_client.py
```diff
@@ -84,11 +84,13 @@ def get_data_points(self, **selection):
         of: item_id, metric_id, region_id, frequency_id, source_id,
         partner_region_id. Additional arguments are allowed and ignored.
         """
-        headers = {'authorization': 'Bearer ' + self.access_token}
-        url = '/'.join(['https:', '', self.api_host, 'v2/data'])
-        params = lib.get_data_call_params(**selection)
-        resp = yield self.get_data(url, headers, params)
-        raise gen.Return(json_decode(resp))
+        data_points = super(BatchClient, self).get_data_points(**selection)
```
tolothy, line 87: i think we are missing a yield here?

nemozen: actually  yield would be if it was calling a coroutine. Here we are calling non-async get_data_points
https://www.tornadoweb.org/en/stable/gen.html#decorators

nemozen: doign it both ways no as discussed offline

api/client/batch_client.py
```diff
@@ -79,22 +79,33 @@ def get_data(self, url, headers, params=None):
...
+    @gen.coroutine
+    def get_data_points(self, **selection):
+        """Get all the data points for a given selection, which is some or all
+        of: item_id, metric_id, region_id, frequency_id, source_id,
+        partner_region_id. Additional arguments are allowed and ignored.
+        """
+        data_points = super(BatchClient, self).get_data_points(**selection)
+        raise gen.Return(data_points)
+
+    def get_df(self):
```
eliaschneider: here we need add option show_revision=True

nemozen: done

tolothy (approved):
tested `batch_async_get_data_points`

PR #152, opened 2019-12-31
Specify unit id in brazil soybean crop model

itsjohnward:
Previously assumed unit would be tonnes per hectare no matter the source. Here, I specified the unit to guarantee consistency.


api/client/samples/crop_models/brazil_soybeans.ipynb
```diff
@@ -42,10 +42,11 @@
      "name": "stdout",
```
JGrinst: <p>This output looks a little more bizarre now. Is this expected due to multiple sources providing yield values?</p><p> </p><p>Same for the first visualized output.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

JGrinst (approved):

api/client/samples/crop_models/brazil_soybeans.ipynb
stephenbernardi: <p>4 sources now, including IBGE</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

stephenbernardi: <p>looks like some strange values are appearing in 2000 or 2001 now</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

stephenbernardi: <p>shouldn't the dataframe be filtered so that there is only a single yield value for each year, to avoid double counting in the regression?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/152/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

itsjohnward: <p>Yes, it's sources reporting very different values. I filtered out PS&amp;D, and that largely cleans it up since they're on a different calendar from the other three sources. I also put a comment in the code explaining that being excluded.</p>

itsjohnward: <p>Updated</p>

stephenbernardi (approved):

api/client/samples/crop_models/brazil_soybeans.ipynb
itsjohnward: If I understand this correctly, I think it's intended to be a weighted regression so we can split the difference on points where we have disagreement between the sources. But I don't see anything in here providing the weights - I think we'd have to give 1/n where n is number of available points for that date. Unless it's intended to double-count to simulate added confidence when multiple sources agree. Not 100% sure the intent, so I'll leave this for @nemozen to comment on

nemozen: right the intention is to use *all* the series for yield when there's more than one, this is an improvement over taking any one series. we are taking advantage of the fact that we have different sources and giving them equal weight.

nemozen:
excluding PS&D is not necessary.  compute_weights is robust to phase shifts

PR #153, opened 2020-01-10
Brazil soybeans update

nemozen:


api/client/samples/crop_models/brazil_soybeans.ipynb
```diff
@@ -92,7 +89,7 @@
...
...BSKNndWRPPbkKuru1K0T4onId7BGBy6wP88CfoJVlXKtMSa/2xLjJ/0hj9A/8tnX8yjWfEUGqyrse8aYPSKShtUNdTKw31j9+Hut/pDHrtZtQUFEgozVP30ZVr/tWqx+wbtNH5lFsj/kEfpHPvtJHgUIw5qq4zWshvKjWF+e5YFMW1fpD3nsaiGtH9I17D+wSKzRnQuBXxpj3umu+3eH/pBH6B/57Cd5NCIyA6u+fTTwf8aa+bTP6A957GrdFhRsN2HV351h7Mnu+qD+kEfoH/nsD3nMwaoS+4PmUUH3tykEGatvcJ/VH/II/SOf/SGPSjXVo1deU0op1b0CPUuqUkqpHkSDglJKKTcNCkoppdw0KCillHLToKCUUspNg4JS7SQile049mkRuaiVY64W30tAKtWtNCgoFXhXAxoUVI+gQUH1SyLyLRH5QkQ2isjf7fWxK0XkARHZJCJrRSTJPna0iHwmIltE5FetXFdE5FERyRaR97GmyXDtu1dE1onIVhF5wj72Iqzpm5+30zJARE4SkdUiskFE3hWRFL/+ZyjViAYF1e+IyASstYVPMcZMx5rm+3KshYHWGmOmAWuAb9unPAL81RgzBcL8S94AAAGbSURBVGtyvJZ8HWt96olYy1g2XlnvUWPMycaYycAA4FxjzDKs9aovt9PSgLV28EXGmJOw1kl+oLN5VqqtunvuI6V6gtOAk4B19hT6A7DWI67DWjcbrKUcz7DfnwJcaL//F9aav77Mx1q0xQHkiciHjfYtEpEfYi0AEwtsA95ocn461pTOK+y0BdN6IFKqy2hQUP2RYC3gfo/HRpG7zIl5Xxx4/n10aj4YEYnAWhIy0xhzWETuAyJ8pG2bMWZuZ+6nVEdp9ZHqjz4ALhKRRAARiRWRUS0c/wnwDfv95a1cew1wqd1GkQIssre7AkCxiEQBjXskVQDR9vtsIEFE5tppCxWRSW3JlFJdQYOC6neMMduBnwLvichmYAXQUmPubcDNIrIFGNbK5V8DdgPbgWeBz+x7lgJPAluBd4F1jc55GvibiGzEqi66CPitiGzCWk+4cbuEUn6ls6QqpZRy05KCUkopN21oVqoDRGQKVk+kxmqNMbMDkR6luopWHymllHLT6iOllFJuGhSUUkq5aVBQSinlpkFBKaWU2/8DWFW+e/ieoa4AAAAASUVORK5CYII=\n",
```
itsjohnward: Could do something like

```py
for source_id, group in yield_df.groupby('source_id'):
    group.plot(x='end_date', y='value', ax=axes, label=model.lookup('sources', source_id)['name'])
```

to plot the sources as separate lines. Looks clearer what's going on as opposed to a single line with multiple y values for the same x

itsjohnward (changes requested):

api/client/samples/crop_models/brazil_soybeans.ipynb
nemozen: done

api/client/samples/crop_models/brazil_soybeans.ipynb
```diff
@@ -8,7 +8,7 @@
     "In this notebook, we walk through the process of building a simple \"crop model\" using data in Gro. The accompanying Gro web app display for [Brazil Soybeans](https://app.gro-intelligence.com/displays/25894) provides an overview of the data series we will use. The crop model we will create is a very basic  model that uses crop-production-weighted NDVI to forecast  yield.\n",
```
itsjohnward, line 8: <p>crop0weighted -&gt; crop-weighted</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/153/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

itsjohnward, line 8: In `compute_crop_weighted_series` we have:

```py
for region in regions:
  entities['region_id'] = region['id']
    for data_series in self.get_data_series(**entities):
      self.add_single_data_series(data_series)
      break
df = self.get_df()
```

which could potentially add in series of varying units. To be safe, we could do something like grab the first unit_id from `df` and then apply `self.convert_unit` to each row to make sure they're consistent before getting the weights. All the resulting series are the same metric_id/item_id, so they should have the same set of allowed units.

<br/>

 _Reply via <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/153/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

nemozen: done

nemozen: actually it doesn't matter here.... since weighting is just applying a bunch of dimensionless constants across values in different regions, it will remain consistent regardless of units as long as units are the same across regions for a given point in time (don't even have to be same units across time or sources)

itsjohnward: Yeah it doesn't matter what the unit is - they just have to be consistent across the regions. But nothing here's guaranteeing that consistency.

You could easily get

```py
{ ..., 'region_id': 1215, 'source_id': 2, 'end_date': '2000-01-01', ''value': 1, 'unit_id': KG },
{ ..., 'region_id': 1216, 'source_id': 14, 'end_date': '2000-01-01', ''value': 1, 'unit_id': TONNES }
```

and then the weights output from `compute_weights` would be off.

nemozen: yes for compute_weights, ```weights = self.compute_weights(...) ``` assumes the weighting_crop_name, weighting_metric_name  series has consistent units across regions. let me add that in a separate PR since that is technically not part of the brazil soybeans

for the item_name and metric_name it doesn't matter since each value is independently has a weight applied to it (naturally the consumer might want to convert to whatever units)

itsjohnward: 👍

itsjohnward (approved):

PR #154, opened 2020-01-10
Update version for show_revision features

eliaschneider:


tolothy (approved):

itsjohnward:
v1.56.2 already has `show_revisions`: https://github.com/gro-intelligence/api-client/blob/v1.56.2/api/client/gro_client.py#L47

This tag doesn't contain any code change.

PR #155, opened 2020-01-10
add __version__ to module

cn1036:


api/version.py
```diff
@@ -0,0 +1 @@
+__version__ = "1.56.2"
```
itsjohnward: I'm thinking we should move toward using https://pypi.org/project/setuptools-scm/ to get the version number from the git tag instead of maintaining the version in two places: in this setup file and in git tags. Right now it's possible to update one and forget to update the other like we saw with v1.56.2:

* https://github.com/gro-intelligence/api-client/pull/149 (version number updated on development branch on 2019-12-23)
* https://github.com/gro-intelligence/api-client/releases/tag/v1.56.2 (version not tagged/released until 2020-01-07)

Would be better to have a single source of truth. Can experiment with that on Monday.

itsjohnward (changes requested):

api/client/lib.py
```diff
@@ -184,6 +186,11 @@ def get_data(url, headers, params=None, logger=None):
     """
     base_log_record = dict(route=url, params=params)
     retry_count = 0
+
+    # append version info
+    headers['api_client_version'] = __version__
```
itsjohnward: Need to update the unit test mocks. Like here:

* https://github.com/gro-intelligence/api-client/blob/3a1cc66e67a20812e7d92ef91824865f927bc971/api/client/lib_test.py#L32

Failing on shippable https://app.shippable.com/github/gro-intelligence/api-client/runs/1773/1/console

itsjohnward: Also unfortunately batch_client has its own version of `get_data()` since it's using a different HTTP library (`tornado` instead of `requests`). So, need to duplicate these headers over there too.

https://github.com/gro-intelligence/api-client/blob/f0003b78e44de2dd0c5d70595f7b30da303182dc/api/client/batch_client.py#L32

itsjohnward (changes requested):

api/version.py
cn1036: done
I hardcoded the name 'gro' in client/lib.py when getting the version but I am not sure if there is any other way to fetch the name defined in setup.py

api/client/lib.py
cn1036: updated

cn1036: done

api/client/batch_client.py
```diff
@@ -37,6 +39,15 @@ def get_data(self, url, headers, params=None):
         base_log_record = dict(route=url, params=params)
         retry_count = 0
         self._logger.debug(url)
+
+        # append version info
+        headers['python_version'] = platform.python_version()
```
muzigao: better use `-` instead of `_`, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

cn1036: updated

nemozen (approved):

api/client/batch_client.py
```diff
@@ -37,6 +39,15 @@ def get_data(self, url, headers, params=None):
         base_log_record = dict(route=url, params=params)
         retry_count = 0
         self._logger.debug(url)
+
+        # append version info
```
nemozen, line 41: minor point; to avoid duplicating these few lies of code you can just have a function 
lib.get_version_info() return a dicts, and then do 

headers.update(lib.get_version_info())

cn1036: done

muzigao (approved):

itsjohnward (approved):

PR #156, opened 2020-01-14
Convert units to default unit in crop weighting series

nemozen:
In case the weighting series happens to be in different units for different regions, we convert tot eh first unit for the selected metric and item

### DIFF
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -221,7 +221,7 @@ def get_data(url, headers, params=None, logger=None):
 
 
 @memoize(maxsize=None)
-def allowed_units(access_token, api_host, metric_id, item_id):
+def get_allowed_units(access_token, api_host, metric_id, item_id):
     url = '/'.join(['https:', '', api_host, 'v2/units/allowed'])
     headers = {'authorization': 'Bearer ' + access_token}
     params = {'metricIds': metric_id}
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -228,7 +228,7 @@ def get_allowed_units(access_token, api_host, metric_id, item_id):
     if item_id:
         params['itemIds'] = item_id
     resp = get_data(url, headers, params)
-    return resp.json()['data']
+    return [unit['id'] for unit in resp.json()['data']]
 
 
 @memoize(maxsize=None)
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -221,6 +221,17 @@ def get_data(url, headers, params=None, logger=None):
 
 
 @memoize(maxsize=None)
+def allowed_units(access_token, api_host, metric_id, item_id):
+    url = '/'.join(['https:', '', api_host, 'v2/units/allowed'])
+    headers = {'authorization': 'Bearer ' + access_token}
+    params = {'metricIds': metric_id}
+    if item_id:
+        params['itemIds'] = item_id
+    resp = get_data(url, headers, params)
+    return resp.json()['data']
+
+
+@memoize(maxsize=None)
 def get_available(access_token, api_host, entity_type):
     url = '/'.join(['https:', '', api_host, 'v2', entity_type])
     headers = {'authorization': 'Bearer ' + access_token}
--- a/api/client/__init__.py
+++ b/api/client/__init__.py
@@ -93,6 +93,10 @@ class Client(object):
     def lookup_unit_abbreviation(self, unit_id):
         return self.lookup('units', unit_id)['abbreviation']
 
+    def allowed_units(self, metric_id, item_id=None):
+        return lib.allowed_units(self.access_token, self.api_host, metric_id,
+                                 item_id)
+
     def get_data_series(self, **selection):
         """Get available data series for the given selections.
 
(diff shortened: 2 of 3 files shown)


api/client/lib.py
```diff
@@ -220,6 +220,17 @@ def get_data(url, headers, params=None, logger=None):
         url, retry_count, data))
 
 
+@memoize(maxsize=None)
+def allowed_units(access_token, api_host, metric_id, item_id):
```
itsjohnward: For the sake of consistency with other functions, I think this should be named `get_allowed_units`. See:

* `get_available`
* `get_data_series`
* `get_data_points`
* `get_geo_centre`
* `get_geojson`
* `get_descendant_regions`

itsjohnward (changes requested):
One comment on the function name but otherwise lgtm on the client side. Tested:

```py
client.allowed_units(860032)
[{'id': 10, 'name': 'kilogram', 'namePlural': 'kilograms', 'abbreviation': 'kg', 'baseUnit': True, 'baseConvFactor': 1, 'convType': 0, 'default': False}, ...

>>> client.allowed_units(860032, 274)
[{'id': 10, 'name': 'kilogram', 'namePlural': 'kilograms', 'abbreviation': 'kg', 'baseUnit': True, 'baseConvFactor': 1, 'convType': 0, 'default': False}, ...

>>> client.allowed_units(1)
{"statusCode":404,"error":"Not Found","message":"No allowed units found (metricIds: 1)"}
Exception: Giving up on https://apistage11201.gro-intelligence.com/v2/units/allowed after 1 tries: <Response [404]>.

>>> client.allowed_units(860032, 99999999999999)
[{'id': 10, 'name': 'kilogram', 'namePlural': 'kilograms', 'abbreviation': 'kg', 'baseUnit': True, 'baseConvFactor': 1, 'convType': 0, 'default': False}, ...
```

Three comments on server-side inconsistencies we should make though before publishing this:

1. `lookup('units', id)` is excluding `'baseUnit'` and `'default'` keys. Should do the same here.
2. `lookup` has `baseConvFactor` as a dict i.e. `{'factor': 1}` instead of on this endpoint it's an int. Also, this int output doesn't work for convtype 1. See for `allowed_units(2540047):

```json
{
            "abbreviation": "°F",
            "baseConvFactor": 0.5555555555555556,
            "baseUnit": false,
            "convType": 1,
            "default": false,
            "id": 37,
            "name": "Fahrenheit",
            "namePlural": "Fahrenheit"
        }
```

it's missing the offset.
3. Item not found should also return a 404, not just metric not found

api/client/lib.py
nemozen: done

nemozen:
> One comment on the function name but otherwise lgtm on the client side. Tested:
> 
> ```python
> client.allowed_units(860032)
> [{'id': 10, 'name': 'kilogram', 'namePlural': 'kilograms', 'abbreviation': 'kg', 'baseUnit': True, 'baseConvFactor': 1, 'convType': 0, 'default': False}, ...
> 
> >>> client.allowed_units(860032, 274)
> [{'id': 10, 'name': 'kilogram', 'namePlural': 'kilograms', 'abbreviation': 'kg', 'baseUnit': True, 'baseConvFactor': 1, 'convType': 0, 'default': False}, ...
> 
> >>> client.allowed_units(1)
> {"statusCode":404,"error":"Not Found","message":"No allowed units found (metricIds: 1)"}
> Exception: Giving up on https://apistage11201.gro-intelligence.com/v2/units/allowed after 1 tries: <Response [404]>.
> 
> >>> client.allowed_units(860032, 99999999999999)
> [{'id': 10, 'name': 'kilogram', 'namePlural': 'kilograms', 'abbreviation': 'kg', 'baseUnit': True, 'baseConvFactor': 1, 'convType': 0, 'default': False}, ...
> ```
> 
> Three comments on server-side inconsistencies we should make though before publishing this:
> 
> 1. `lookup('units', id)` is excluding `'baseUnit'` and `'default'` keys. Should do the same here.
> 2. `lookup` has `baseConvFactor` as a dict i.e. `{'factor': 1}` instead of on this endpoint it's an int. Also, this int output doesn't work for convtype 1. See for `allowed_units(2540047):
> 
> ```json
> {
>             "abbreviation": "°F",
>             "baseConvFactor": 0.5555555555555556,
>             "baseUnit": false,
>             "convType": 1,
>             "default": false,
>             "id": 37,
>             "name": "Fahrenheit",
>             "namePlural": "Fahrenheit"
>         }
> ```
> 
> it's missing the offset.

good point... to keep it simple, I made get_allowed_units just return of list of unit ids.

> 3. Item not found should also return a 404, not just metric not found

Hmm, actually we  don't want to  fail in that case since there's is a valid response. (  the unit that works for the metric in general always works for all items,  it's logically correct to say units of mass for foo is Kg even if Foo doesn't exist, if it did, its mass would be in Kg :)  In general for http requets the convention is to  ignore harmless get params if there is a valid response http://www.google.com/search?q=foo&nonexistentparam=bar will just ignore the second param

itsjohnward (approved):

PR #159, opened 2020-01-22
CLEWS-22325 get data points (omit historical regions)

seungatGro:
Adding option to get_data_points function to include historical region. This will be true by default, and users can have the option to exclude from response.

get_data_points:
```
selections = {}
selections['region_id'] = 14
selections['metric_id'] = 860032
selections['item_id'] = 5074
selections['source_id'] = 14
selections['frequency_id'] = 9
client.get_data_points(include_historical=False, **selections)
```

api response flags series metadata for historical region, api client has the option to filter these

### DIFF
--- a/api/client/__init__.py
+++ b/api/client/__init__.py
@@ -144,8 +144,8 @@ class Client(object):
         return lib.get_data_series(self.access_token, self.api_host, **selection)
 
 
-    def get_data_points(self, **selection):
-        return lib.get_data_points(self.access_token, self.api_host, **selection)
+    def get_data_points(self, include_historical=True, **selection):
+        return lib.get_data_points(self.access_token, self.api_host, include_historical=include_historical, **selection)
 
 
     def search(self, entity_type, search_terms):
--- a/api/client/__init__.py
+++ b/api/client/__init__.py
@@ -144,8 +144,8 @@ class Client(object):
         return lib.get_data_series(self.access_token, self.api_host, **selection)
 
 
-    def get_data_points(self, include_historical=True, **selection):
-        return lib.get_data_points(self.access_token, self.api_host, include_historical=include_historical, **selection)
+    def get_data_points(self, **selection):
+        return lib.get_data_points(self.access_token, self.api_host, **selection)
 
 
     def search(self, entity_type, search_terms):
--- a/api/client/batch_client.py
+++ b/api/client/batch_client.py
@@ -83,12 +83,14 @@ class BatchClient(GroClient):
                         url, retry_count, e.response.error))
 
     @gen.coroutine
-    def get_data_points(self, **selection):
+    def get_data_points(self, include_historical=True, **selection):
         """Get all the data points for a given selection, which is some or all
         of: item_id, metric_id, region_id, frequency_id, source_id,
         partner_region_id. Additional arguments are allowed and ignored.
         """
-        data_points = super(BatchClient, self).get_data_points(**selection)
+        data_points = super(BatchClient, self).get_data_points(
+            include_historical=include_historical, 
+            **selection)
         raise gen.Return(data_points)
 
     def get_df(self, show_revisions=True):
@@ -102,7 +104,7 @@ class BatchClient(GroClient):
     # TODO: deprecate  the following  two methods, standardize  on one
     # approach with get_data_points and get_df
     @gen.coroutine
-    def get_data_points_generator(self, **selection):
+    def get_data_points_generator(self, include_historical=True, **selection):
         headers = {'authorization': 'Bearer ' + self.access_token}
         url = '/'.join(['https:', '', self.api_host, 'v2/data'])
         params = lib.get_data_call_params(**selection)
@@ -110,10 +112,12 @@ class BatchClient(GroClient):
         raise gen.Return(json_decode(resp))
 
     def batch_async_get_data_points(self, batched_args, output_list=None,
-                                    map_result=None):
+                                    map_result=None, include_historical=True):
+        for args in batched_args:
+            args['include_historical'] = include_historical
         batch_async_series_list = self.batch_async_queue(
             self.get_data_points_generator, batched_args, output_list, map_result)
-        return [lib.list_of_series_to_single_series(series_list) for series_list in batch_async_series_list]
+        return [lib.list_of_series_to_single_series(series_list, False, include_historical) for series_list in batch_async_series_list]
 
     @gen.coroutine
     def async_rank_series_by_source(self, **selection):
--- a/api/client/batch_client.py
+++ b/api/client/batch_client.py
@@ -83,14 +83,12 @@ class BatchClient(GroClient):
                         url, retry_count, e.response.error))
 
     @gen.coroutine
-    def get_data_points(self, include_historical=True, **selection):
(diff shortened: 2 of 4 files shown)


api/client/__init__.py
```diff
@@ -144,8 +144,8 @@ def get_data_series(self, **selection):
         return lib.get_data_series(self.access_token, self.api_host, **selection)
 
 
-    def get_data_points(self, **selection):
-        return lib.get_data_points(self.access_token, self.api_host, **selection)
+    def get_data_points(self, include_historical=True, **selection):
```
itsjohnward: Should `include_historical` should be one of the `selection`s, like `insert_null` or `show_revisions`? It just happens to be a client-side "selection" that'd get filtered out from the server request in `get_data_call_params()`, like how `unit_id` is

api/client/batch_client.py
```diff
@@ -102,18 +104,20 @@ def get_df(self, show_revisions=True):
...
         headers = {'authorization': 'Bearer ' + self.access_token}
         url = '/'.join(['https:', '', self.api_host, 'v2/data'])
         params = lib.get_data_call_params(**selection)
         resp = yield self.get_data(url, headers, params)
         raise gen.Return(json_decode(resp))
 
     def batch_async_get_data_points(self, batched_args, output_list=None,
-                                    map_result=None):
+                                    map_result=None, include_historical=True):
+        for args in batched_args:
```
itsjohnward: Can put the requirement on the caller to include it in each of the `batched_args` like

```py
batched_selections = [{
    'metric_id': 2100031,
    'item_id': 2039,
    'region_id': region_batch,
    'source_id': 35,
    'frequency_id': 1,
    'insert_null': True,
    'include_historical': False
} for region_batch in batch(batch_size)]
client.batch_async_get_data_points(batched_selections)
````

instead of calling `batch_async_get_data_points` with an additional arg. And this way if you really wanted to, it could vary from selection to selection - you might not want the same option on all of them

**Review by itsjohnward — CHANGES_REQUESTED** [review]

File: api/client/batch_client.py (hunk 2 above)
Comment (seungatGro) [reply api/client/batch_client.py]: discussed offline. previously had wanted to keep it separate from entities, but these will be removed in respective lib `get_data_call_params` fn and allows for batch call to have an option per selection. pushed changes to remove as an explicit option and keep as a selection

File: api/client/__init__.py (hunk 1 above)
Comment (seungatGro) [reply api/client/__init__.py]: commented below on `BatchClient. batch_async_get_data_points`

File: api/client/batch_client.py [hunk 3]
```diff
@@ -104,7 +104,7 @@ def get_df(self, show_revisions=True):
     @gen.coroutine
     def get_data_points_generator(self, **selection):
         headers = {'authorization': 'Bearer ' + self.access_token}
-        url = '/'.join(['https:', '', self.api_host, 'v2/data'])
+        url = '/'.join(['http:', '', self.api_host, 'v2/data'])
```
itsjohnward: TODO: revert before merge

api/client/lib.py
```diff
@@ -513,6 +516,8 @@ def list_of_series_to_single_series(series_list, add_belongs_to=False):
                 'frequency_id': series['series'].get('frequencyId', None)
                 # 'source_id': series['series'].get('sourceId', None), TODO: add source to output
             }
+            if series_metadata:
+                formatted_point['metadata'] = series_metadata
```
itsjohnward: This would duplicate the series-level metadata onto every single point and looks like point-level metadata but isn't. Not sure if or how we should differentiate series-level vs. point-level. Not important to add this metadata at the moment, right? Maybe just exclude this bit for now until we have a clear spec on what that should look like.

api/client/lib.py
```diff
@@ -487,6 +487,9 @@ def list_of_series_to_single_series(series_list, add_belongs_to=False):
     for series in series_list:
         if not (isinstance(series, dict) and isinstance(series.get('data', []), list)):
             continue
+        series_metadata = series.get('series', {}).get('metadata', {})
+        if not include_historical and series_metadata.get('includesHistoricalRegion', False):
```
itsjohnward: There's also `'includesHistoricalPartnerRegion'`. Going to check for that as well?

Ex: `client.get_data_points(metric_id=11078, item_id=1, region_id=1079, partner_region_id=14, frequency_id=9, source_id=1)`

Also, I am noticing that the `'belongsTo'` series is including the metadata as well. Is that intended?

api/client/lib.py
```diff
@@ -513,6 +516,8 @@ def list_of_series_to_single_series(series_list, add_belongs_to=False):
                 'frequency_id': series['series'].get('frequencyId', None)
                 # 'source_id': series['series'].get('sourceId', None), TODO: add source to output
             }
+            if series_metadata:
```
itsjohnward: If this metadata's included, it will break the pandas `get_df()` function again like we had with the `'belongsTo'` property https://github.com/gro-intelligence/api-client/pull/147/files. Can't hash the dict values. So, another reason to exclude metadata for now like I said in my other comment.

itsjohnward (changes requested):

api/client/lib.py
seungatGro: good point, removing from point level formatting

api/client/lib.py
seungatGro: this is intended. if the user requests data for a series selection, we expand to contained series, and i purposely mapped this back to the user selected series. see series-availability-utils addToAvailableSeriesSet  https://bitbucket.org/grointelligence/gro/src/3aa37b417e427aaa7b69dafbe70586a255e95585/api/lib/helpers/series-availability-utils.js#lines-368

seungatGro: done. added check for historical partner, tested with above selections

api/client/lib.py
seungatGro: removed metadata from point

itsjohnward (approved):

itsjohnward:
Tested GroClient `get_data_points()` and `get_df()` as well as BatchClient `batch_async_get_data_points()`, with and without historical regions:

```py
    series_with_historical = {
        'metric_id': 11078,
        'item_id': 1,
        'region_id': 1079,
        'partner_region_id': 14,
        'frequency_id': 9,
        'source_id': 1
    }

    series_without_historical = copy.deepcopy(series_with_historical)
    series_without_historical['include_historical'] = False

    for series, msg in [
        (series_with_historical, 'with historical'),
        (series_without_historical, 'without historical')
    ]:
        print('\n{}'.format(msg))

        client._logger.setLevel(logging.ERROR)
        print('get_data_points: {}'.format(len(client.get_data_points(**series))))

        client._logger.setLevel(logging.ERROR)
        client.add_single_data_series(series)
        print('get_df: {}'.format(len(client.get_df())))

        batch_client = BatchClient(API_HOST, ACCESS_TOKEN)
        batch_client._logger.setLevel(logging.ERROR)
        print('batch_async_get_data_points: {}'.format(len(batch_client.batch_async_get_data_points([series])[0])))
```

```sh
$ python test_historical.py 

with historical
get_data_points: 25
get_df: 25
batch_async_get_data_points: 25

without historical
get_data_points: 23
get_df: 23
batch_async_get_data_points: 23
```

PR #161, opened 2020-01-23
CLEWS-19840 Lookup lists of entities at a time

itsjohnward:
Ready now. Tested:

```py
>>> lib.lookup('regions', 1215)
{'1215': {'aliases': ['US', 'United States of America', 'America', 'USA'], 'contains': [13100, 13061, ...], 'definition': None, 'historical': False, 'id': 1215, 'latitude': 39.8333, 'level': 3, 'longitude': -98.5855, 'name': 'United States', 'belongsTo': [0, 15, ...]}}

>>> client.lookup('regions', 1215)
{'aliases': ['US', 'United States of America', 'America', 'USA'], 'contains': [13100, 13061, ...], 'definition': None, 'historical': False, 'id': 1215, 'latitude': 39.8333, 'level': 3, 'longitude': -98.5855, 'name': 'United States', 'belongsTo': [0, 15, ...]}

>>> client.lookup('regions', [1215,1216])
{ '1215': {'aliases': ['US', 'United States of America', 'America', 'USA'], 'contains': [13100, 13061, ...], 'definition': None, 'historical': False, 'id': 1215, 'latitude': 39.8333, 'level': 3, 'longitude': -98.5855, 'name': 'United States', 'belongsTo': [0, 15, ...]}, 
  '1216': {'aliases': ['URY'], 'contains': [13114, 13109, ...], 'definition': None, 'historical': False, 'id': 1216, 'latitude': -32.794, 'level': 3, 'longitude': -56.0178, 'name': 'Uruguay', 'belongsTo': [0, 16, ...]} }
```

NOTE: the output of `lib.lookup()` changed to support multiple entities returned, but I put a check in Client to translate `client.lookup()` back to the old format if an integer's given, rather than a list.


api/client/constants.py
```diff
@@ -0,0 +1,37 @@
+"""Constants about the Gro ontology that can be imported and re-used anywhere."""
+
+REGION_LEVELS = {
```
seungatGro, line 3: minor. same constant in lib, referred to in client

api/client/lib.py
```diff
@@ -296,11 +296,11 @@ def list_available(access_token, api_host, selected_entities):
         raise Exception(resp.text)
 
 
-@memoize(maxsize=None)
-def lookup(access_token, api_host, entity_type, entity_id):
-    url = '/'.join(['https:', '', api_host, 'v2', entity_type, str(entity_id)])
+def lookup(access_token, api_host, entity_type, entity_ids):
+    url = '/'.join(['https:', '', api_host, 'v2', entity_type])
```
seungatGro, line 245: i'd suggest backward compatibility here too for older clients as @itsjohnward brought up in stand up..

api/client/utils.py
```diff
@@ -0,0 +1,99 @@
+"""Miscellaneous utility functions for operations not specific to the API client."""
```
seungatGro, line 1: is this intentionally committed? not seeing any callers

seungatGro (changes requested):

api/client/utils.py
alexsherman: Similar to your comment about the constants above, these seem to currently exist in `lib` but haven't been replaced. Is this something you're still intending to do John?

alexsherman (commented):
The changes look good to me with the exception of the duplication between lib, constants and utils.

I'm not sure what you guys decided about lib's backwards compatibility in the sync. If it needs to be fully backwards compatible, then as you mentioned, some of the stuff in client lookup will have to be shifted. But the logic itself looks fine.

api/client/constants.py
itsjohnward: Whoops. Stuff got added back in when resolving a merge conflict. Removed. Now it's importing from the constants file

api/client/utils.py
itsjohnward: Whoops. Stuff got added back in when resolving a merge conflict. Removed. Now it's importing from the utils file

api/client/lib.py
itsjohnward: I just went ahead and moved the translation bit into here since if we *do* want lib to always be backwards compatible, then it's necessary, but if we decide it *doesn't* need to be then we can always just refactor it back later. So, erred on the safe side

seungatGro (approved):
LGTM. tested with list and singular

alexsherman (approved):

api/client/lib.py
```diff
@@ -528,13 +458,15 @@ def list_of_series_to_single_series(series_list, add_belongs_to=False, include_h
...
+        has_historical_regions = (series_metadata.get('includesHistoricalRegion', False) or
+                                  series_metadata.get('includesHistoricalPartnerRegion', False))
+        if not include_historical and has_historical_regions:
             continue
         # All the belongsTo keys are in camelCase. Convert them to snake_case.
         # Only need to do this once per series, so do this outside of the list
         # comprehension and save to a variable to avoid duplicate work:
-        belongs_to = camel_to_snake_dict(series.get('series', {}).get('belongsTo', {}))
+        belongs_to = dict_reformat_keys(series.get('series', {}).get('belongsTo', {}),
+                                        str_camel_to_snake)
```
seungatGro, line 469: getting error here `str_camel_to_snake() takes 1 positional argument but 2 were given`

itsjohnward: Should be fixed now. Fantastic save - there was a paren out of place. Instead of

```py
dict(format_func((key), value) for key, value in obj.items())
```

it was supposed to be

```py
dict((format_func(key), value) for key, value in obj.items())
```

I was very surprised that the unit test here https://github.com/gro-intelligence/api-client/blob/development/api/client/lib_test.py#L149 didn't catch that, but it's just because the `series` object in that test has no keys. So this loop runs 0 times and that function's never called.

I have now:

1. re-factored that line to not have so many parens so it's more readable https://github.com/gro-intelligence/api-client/pull/161/commits/b5478d0519864dd14c477382da79331d995ba16f
2. added the doctests to shippable. The test was failing for that function, but shippable's never been automatically running the doctests until now https://github.com/gro-intelligence/api-client/pull/161/commits/637e43284b003d8c3ef2039729b334d99320bc56

seungatGro:
changed LGTM. tested batching, calling str formatters

seungatGro (approved):

PR #166, opened 2020-02-13
CLEWS-24615: API client errors handling

waracci:
Description
- These are the errors that are handled by the added exception **400, 401, 404, 500**
- The current implementation returns a response as shown below 👇 

> _API call failed with the error message:_
> Giving up on https://api.gro-intelligence.com/sample_unit/ after 1 tries:

- Seeing as it is difficult to parse the message ☝️ the error message will include the following info in the response body

1. Status Code
2. Retry Count
3. Response
4. Message
5. Url

### DIFF
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -244,7 +244,7 @@ def get_data(url, headers, params=None, logger=None):
             logger.warning('Redirecting {} to {}'.format(params, new_params), extra=log_record)
             params = new_params
         elif data.status_code in [400, 401, 404, 500]:
-            raise APIError(data, retry_count, url)
+            break
         else:
             logger.error('{}'.format(data), extra=log_record)
     raise APIError(data, retry_count, url)
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -34,6 +34,18 @@ REGION_LEVELS = {
 }
 
 
+class APIError(Exception):
+    def __init__(self, response, retry_count, url):
+        self.response = response
+        self.retry_count = retry_count
+        self.url = url
+        self.status_code = self.response.status_code
+        self.text = self.response.text
+        self.message = 'Giving up on {} after {} tries: {}.'.format(self.url,
+                                                                    self.retry_count,
+                                                                    self.response)
+
+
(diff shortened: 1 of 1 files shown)


api/client/lib.py
```diff
@@ -232,7 +244,10 @@ def get_data(url, headers, params=None, logger=None):
...
             params = new_params
         elif data.status_code in [400, 401, 404, 500]:
-            break
+            try:
+                raise APIError(data, retry_count, url)
+            except APIError as error:
+                return error
         else:
             logger.error('{}'.format(data), extra=log_record)
     raise Exception('Giving up on {} after {} tries: {}.'.format(
```
itsjohnward: Let's also raise an APIError here. This one's if you're timing out repeatedly and it gives up, so `data` in this case would likely be a 503 or 504 error.

itsjohnward (changes requested):

api/client/lib.py
waracci: Done!

itsjohnward:
Hey @waracci FYI I just added some message parsing on the error responses to something I think is a bit more useful:

Test script:

```py
try:
    client.get_data_points(region_id=19)
except Exception as e:
    print('e {}'.format(e))
    print('response {}'.format(e.response))
    print('retry_count {}'.format(e.retry_count))
    print('url {}'.format(e.url))
    print('status_code {}'.format(e.status_code))
    print('message {}'.format(e.message))
```

Output changes from:

```py
response <Response [400]>
retry_count 1
url https://api.gro-intelligence.com/v2/data
status_code 400
message Giving up on https://api.gro-intelligence.com/v2/data after 1 tries: <Response [400]>.
```

to:

```py
response <Response [400]>
retry_count 1
url https://api.gro-intelligence.com/v2/data
status_code 400
message Bad Request: child "metricId" fails because ["metricId" is required]
```

Also I added `params` to the error object. Use case would be if you're looping through many different requests to the same endpoint and catching the errors as you go, you can access the `params` property to retroactively see which inputs resulted in the error.

itsjohnward (approved):

PR #171, opened 2020-02-25
Direct git to ignore commonly untracked Python and Jetbrains files.

MichaelWakahe:
This is a .gitignore file with files that should not be tracked by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,123 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+pip-wheel-metadata/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.nox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+*.py,cover
+.hypothesis/
+.pytest_cache/
+cover/
+
+# Sphinx documentation
+docs/_build/
+
+# PyBuilder
+target/
+
+# Jupyter Notebook
+.ipynb_checkpoints
+
+# IPython
+profile_default/
+ipython_config.py
+
+# pyenv
+#   For a library or package, you might want to ignore these files since the code is
+#   intended to run in multiple environments; otherwise, check them in:
+# .python-version
+
+# pipenv
+#   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in version control.
+#   However, in case of collaboration, if having platform-specific dependencies or dependencies
+#   having no cross-platform support, pipenv may install dependencies that don't work, or not
+#   install all needed dependencies.
+#Pipfile.lock
+
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow
+__pypackages__/
+
+# Celery stuff
+celerybeat-schedule
+celerybeat.pid
+
+# SageMath parsed files
+*.sage.py
+
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+
+# Spyder project settings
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -0,0 +1,123 @@
...
+nosetests.xml
+coverage.xml
+*.cover
+*.py,cover
+.hypothesis/
+.pytest_cache/
+cover/
+
+# Sphinx documentation
+docs/_build/
```
itsjohnward, line 56: We have this in `docs/.gitignore` currently. Can remove one or the other

.gitignore
```diff
@@ -0,0 +1,123 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
```
itsjohnward, line 3: Ignoring `.pyc` in api/.gitignore currently. Should be able to remove that file

.gitignore
```diff
@@ -0,0 +1,123 @@
...
+# Pyre type checker
+.pyre/
+
+# pytype static type analyzer
+.pytype/
+
+
+# IntelliJ
+*.iml
+.idea/
```
itsjohnward, line 123: Might as well add these for VSCode as well:

```py
.vscode/*
*.code-workspace
```

.gitignore
```diff
@@ -0,0 +1,123 @@
...
+
+# Spyder project settings
+.spyderproject
+.spyproject
+
+# Rope project settings
+.ropeproject
+
+# mkdocs documentation
+/site
```
itsjohnward, line 107: We are using sphinx, rather than mkdocs, so this one can be excluded

.gitignore
```diff
@@ -0,0 +1,123 @@
...
+# mypy
+.mypy_cache/
+.dmypy.json
+dmypy.json
+
+# Pyre type checker
+.pyre/
+
+# pytype static type analyzer
+.pytype/
```
itsjohnward, line 118: I would like to add mypy at some point, but currently we're not using any of these type checkers (mypy, pyre, pytype). So I think we can exclude them until it's relevant

.gitignore
```diff
@@ -0,0 +1,123 @@
...
+ENV/
+env.bak/
+venv.bak/
+
+# Spyder project settings
+.spyderproject
+.spyproject
+
+# Rope project settings
+.ropeproject
```
itsjohnward, line 104: Not using rope currently

.gitignore
```diff
@@ -0,0 +1,123 @@
...
+#   having no cross-platform support, pipenv may install dependencies that don't work, or not
+#   install all needed dependencies.
+#Pipfile.lock
+
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow
+__pypackages__/
+
+# Celery stuff
+celerybeat-schedule
+celerybeat.pid
```
itsjohnward, line 85: Nothing celery-related in this repo. Can use the API client in a celery task, but scheduling and all that would not go in this package. Can exclude

.gitignore
```diff
@@ -0,0 +1,123 @@
...
+#   For a library or package, you might want to ignore these files since the code is
+#   intended to run in multiple environments; otherwise, check them in:
+# .python-version
+
+# pipenv
+#   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in version control.
+#   However, in case of collaboration, if having platform-specific dependencies or dependencies
+#   having no cross-platform support, pipenv may install dependencies that don't work, or not
+#   install all needed dependencies.
+#Pipfile.lock
```
itsjohnward, line 78: We can uncomment to ignore Pipfile.lock.
You want it committed in an *application*, like what we deploy to our servers, but not in an installable package. Here, we want requirements to be as permissive as possible so it doesn't cause unnecessary conflicts in installers' environments.

.gitignore
```diff
@@ -0,0 +1,123 @@
...
+.ipynb_checkpoints
+
+# IPython
+profile_default/
+ipython_config.py
+
+# pyenv
+#   For a library or package, you might want to ignore these files since the code is
+#   intended to run in multiple environments; otherwise, check them in:
+# .python-version
```
itsjohnward, line 71: Can uncomment to ignore. We support many different python versions - shouldn't need to specify a specific one with pyenv

.gitignore
```diff
@@ -0,0 +1,123 @@
...
+*.py,cover
+.hypothesis/
+.pytest_cache/
+cover/
+
+# Sphinx documentation
+docs/_build/
+
+# PyBuilder
+target/
```
itsjohnward, line 59: Not using PyBuilder - can remove

itsjohnward (changes requested):
Looks all right, but I think it's somewhat overkill - left some comments on irrelevant lines so we can slim it down a bit to what's really necessary.

Also, let's consolidate `api/.gitignore`, `docs/.gitignore`, and `r-client/.gitignore` up into this one

jli:
the major parts of this were added in #264, thanks @MichaelWakahe !

PR #172, opened 2020-02-28
CLEWS-19840 Part 2: Use batch lookup in derivative methods

itsjohnward:
After https://github.com/gro-intelligence/api-client/pull/161 is merged, derivative methods using lookup can use batching to reduce their number of requests.

Here, `search_and_lookup`, `lookup_belongs`, and `get_geo_centre` all have their `lookup()` calls replaced with batched versions.


api/client/lib.py
```diff
@@ -244,16 +245,23 @@ def list_available(access_token, api_host, selected_entities):
 def lookup(access_token, api_host, entity_type, entity_ids):
     url = '/'.join(['https:', '', api_host, 'v2', entity_type])
     headers = {'authorization': 'Bearer ' + access_token}
-    params = {'ids': str(entity_ids)}
-    resp = get_data(url, headers, params)
-    try:
-        # If an integer is given, return only the dict with that id
-        if isinstance(entity_ids, int):
+    # If an integer is given, return only the dict with that id
+    if isinstance(entity_ids, int):
```
seungatGro, line 253: should update with dev branch for similar error with df

api/client/lib.py
```diff
@@ -542,18 +550,18 @@ def search(access_token, api_host, entity_type, search_terms):
...
 
 def search_and_lookup(access_token, api_host, entity_type, search_terms, num_results=10):
-    search_results = search(access_token, api_host, entity_type, search_terms)
-    for result in search_results[:num_results]:
-        yield lookup(access_token, api_host, entity_type, result['id'])
+    search_results = search(access_token, api_host, entity_type, search_terms)[:num_results]
+    search_result_ids = [result['id'] for result in search_results]
+    search_result_details = lookup(access_token, api_host, entity_type, search_result_ids)
+    for search_result_id in search_result_ids:
+        yield search_result_details[str(search_result_id)]
```
seungatGro, line 573: do we need to use yield here? lookup will return all search result ids

seungatGro: discussed offline. replacing here with return is a breaking change for callers

seungatGro (changes requested):
only change req is merging dev for lookup

api/client/lib.py
itsjohnward: done

seungatGro (approved):
LGTM

PR #175, opened 2020-03-02
CLEWS-24002 / CLEWS-25014 Retry 500 error codes and add exponential backoff to non-429 responses

itsjohnward:
A paraphrased summary of error codes we use and how I think get_data should respond to them:

Status code | Meaning | Retry
------------ | --------- | -----
301 | Data's been migrated. Here are the updated ids. | yes, with redirect
400 | User provided bad inputs | no
401 | Authentication's not right. Check your token | no
402 | Payment required. Your account's out of points, or you're trying to access data you need an upgrade for | no
404 | User requested something that doesn't exist | no
429 | Too many requests. Rate limited | yes
500 | Server error | yes
503 | API server timed out | yes
504 | An upstream service (db, cache, search, etc.) timed out | yes

For the last two, if a large list of things was requested, we could potentially automatically split the list into two smaller requests to help the request go through. Future work.

Changes in this PR:

* Automatically retry 500 errors.
* Previously, only 429 rate limit responses had exponential backoff. In this PR, I added it to 500, 503, and 504, all of which can benefit from minimizing thrashing on errors.
* Add official support for "payment required" and "partial content" responses. Previously, both would throw errors.


muzigao (approved):
LGTM

api/client/lib.py
```diff
@@ -236,26 +236,26 @@ def get_data(url, headers, params=None, logger=None):
...
+        if response.status_code == 301:
             new_params = redirect(params, response.json()['data'][0])
             logger.warning('Redirecting {} to {}'.format(params, new_params), extra=log_record)
             params = new_params
-        elif response.status_code in [400, 401, 404, 500]:
-            break
         else:
-            logger.error('{}'.format(response), extra=log_record)
+            logger.warning('{}'.format(response), extra=log_record)
+            time.sleep(2 ** retry_count)  # Exponential backoff before retrying
```
nemozen: hmm not sure we want exponential backoff in all cases, immediate retry is. why not keep 429 as a special case?

nemozen (commented):
lgtm just one comment

api/client/lib.py
itsjohnward: It's a trade-off, right? Immediate retries maximize potential throughput, but adding backoffs increases the likelihood that a request will succeed eventually without the user needing to reinvent the wheel with their own retrying.

If the issue's limited to a single load-balanced node or one random dropped connection, immediate retry would be more optimal, but sometimes a process might be down momentarily and immediately responding errors - good to give it a chance to start back up instead of getting 5 rapid errors in a row and crashing.

A minor optimization: we could get the best of both worlds by doing one immediate retry and then only backing off on subsequent ones, so at least it doesn't add that unnecessary second of delay to a one-off error.

itsjohnward: ^ added immediate retry for first failure

nemozen (approved):

tolothy (approved):
:+1:

PR #177, opened 2020-03-05
CLEWS-25750 CLEWS-25400 Retry BatchClient on other types of errors

itsjohnward:
Things resolved in this PR:

* Was only catching HTTPError codes before (non-200 status code responses). Now we're catching other exceptions, like broken sockets, and retrying those.
* Previously when one of the above non-HTTPError exceptions would occur, the Tornado event loop would never complete. Now, we're always either raising a response or an error.
* Now the BatchErrors will be returned by default, so the user can handle them at the end or via map_result.

Testing script:

```py
selections = [
    {'metric_id': 570001, 'item_id': 274, 'region_id': r['id'], 'source_id': 25, 'frequency_id': 9}
    for r in client.get_descendant_regions(1215, descendant_level=4, include_details=False)
]

try:
    for e in client.batch_async_get_data_points(selections):
        print(e.message)
except Exception as e:
    print('EXCEPTION CAUGHT:', e)

try:
    def ignore_nulls(idx, query, response, accumulator):
        accumulator[idx] = None if isinstance(response, Exception) else response
        return accumulator
    for result in client.batch_async_get_data_points(selections, map_result=ignore_nulls):
        print(result)
except Exception as e:
    print('EXCEPTION CAUGHT:', e)
```

In the first loop you get BatchError objects, and their message is printed. So it appears like:

```py
...
WARNING:api.client.lib:HTTP 500: Internal Server Error
HTTP 500: Internal Server Error
WARNING:api.client.lib:HTTP 500: Internal Server Error
HTTP 500: Internal Server Error
WARNING:api.client.lib:HTTP 500: Internal Server Error
HTTP 500: Internal Server Error
WARNING:api.client.lib:HTTP 500: Internal Server Error
Internal Server Error: An internal server error occurred
Internal Server Error: An internal server error occurred
Internal Server Error: An internal server error occurred
Internal Server Error: An internal server error occurred
Internal Server Error: An internal server error occurred
...
```

In the second block, map_result is converting the 500 errors into Nones, so it looks like this:

```py
...
WARNING:api.client.lib:HTTP 500: Internal Server Error
HTTP 500: Internal Server Error
WARNING:api.client.lib:HTTP 500: Internal Server Error
HTTP 500: Internal Server Error
WARNING:api.client.lib:HTTP 500: Internal Server Error
HTTP 500: Internal Server Error
WARNING:api.client.lib:HTTP 500: Internal Server Error
None
None
None
None
None
...
```

Or, if you want to exit execution immediately when one of the batched queries is returning an error, you can raise it in the map_result function to break the IOLoop:

```py
try:
    def raise_exception(idx, query, response, accumulator):
        if isinstance(response, Exception):
            print('{} {} response is an Exception: [{}] {}'.format(response.url, response.params, response.status_code, response.message))
            raise response
        accumulator[idx] = response
        return accumulator

    for result in client.batch_async_get_data_points(selections, map_result=raise_exception):
        print(result)
except Exception as e:
    print('EXCEPTION CAUGHT:', e)
```

which looks like

```py
HTTP 500: Internal Server Error
WARNING:api.client.lib:HTTP 500: Internal Server Error
http://localhost:5000/v2/data {'metricId': 570001, 'itemId': 274, 'regionId': 13051, 'sourceId': 25, 'frequencyId': 9, 'responseType': 'list_of_series'} response is an Exception: [500] Internal Server Error: An internal server error occurred
EXCEPTION CAUGHT: Operation timed out after None seconds
```

### DIFF
--- a/api/client/batch_client.py
+++ b/api/client/batch_client.py
@@ -116,7 +116,10 @@ class BatchClient(GroClient):
                                                  hasattr(e.response, 'error')) else e
                 log_request(start_time, retry_count, error_msg, status_code)
                 if status_code in [429, 500, 503, 504]:
-                    time.sleep(2 ** retry_count)  # Exponential backoff before retrying.
+                    # First retry is immediate.
+                    # After that, exponential backoff before retrying.
+                    if retry_count > 1:
+                        time.sleep(2 ** retry_count)
                     continue
                 elif status_code in [400, 401, 402, 404]:
                     break  # Do not retry. Go right to raising an Exception.
--- a/api/client/batch_client.py
+++ b/api/client/batch_client.py
@@ -297,19 +297,25 @@ class BatchClient(GroClient):
         def consumer():
             """Execute func on all items in queue asynchronously."""
             while q.qsize():
-                idx, item = q.get().result()
-                self._logger.debug('Doing work on {}'.format(idx))
-                if type(item) is dict:
-                    # Assume that dict types should be unpacked as kwargs
-                    result = yield func(**item)
-                elif type(item) is list:
-                    # Assume that list types should be unpacked as positional args
-                    result = yield func(*item)
-                else:
-                    result = yield func(item)
-                output_data['result'] = map_result(idx, item, result, output_data['result'])
-                self._logger.debug('Done with {}'.format(idx))
-                q.task_done()
+                try:
+                    idx, item = q.get().result()
+                    self._logger.debug('Doing work on {}'.format(idx))
+                    if type(item) is dict:
+                        # Assume that dict types should be unpacked as kwargs
+                        result = yield func(**item)
+                    elif type(item) is list:
+                        # Assume that list types should be unpacked as positional args
+                        result = yield func(*item)
+                    else:
+                        result = yield func(item)
+                    output_data['result'] = map_result(idx, item, result, output_data['result'])
+                    self._logger.debug('Done with {}'.format(idx))
+                    q.task_done()
+                except Exception:
+                    # Cease processing
+                    # IOLoop raises "Operation timed out after None seconds"
+                    IOLoop.current().stop()
+                    IOLoop.current().close()
 
         def producer():
             """Immediately enqueue the whole batch of requests."""
--- a/api/client/batch_client.py
+++ b/api/client/batch_client.py
@@ -72,13 +72,14 @@ class BatchClient(GroClient):
 
         Assigns headers and builds in retries and logging.
         """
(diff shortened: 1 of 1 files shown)


api/client/batch_client.py
```diff
@@ -30,61 +55,81 @@ def __init__(self, api_host, access_token):
...
         Assigns headers and builds in retries and logging.
         """
-        base_log_record = dict(route=url, params=params)
         retry_count = 0
         self._logger.debug(url)
 
         # append version info
         headers.update(lib.get_version_info())
 
         while retry_count < cfg.MAX_RETRIES:
```
nemozen: oh actually this should be <= since if you se MAX_RETRIES = 1 you expect it to be one retry in addition to the original try

api/client/batch_client.py
```diff
@@ -30,61 +55,81 @@ def __init__(self, api_host, access_token):
...
+                    # After that, exponential backoff before retrying.
+                    if retry_count > 1:
+                        time.sleep(2 ** retry_count)
+                    continue
+                elif status_code in [400, 401, 402, 404]:
+                    break  # Do not retry. Go right to raising an Exception.
+
+            # Request was successful
+            log_request(start_time, retry_count, 'OK', status_code)
+            raise gen.Return(json_decode(response.body) if hasattr(response, 'body') else None)
```
nemozen, line 130: why json_decode?

nemozen (commented):
lgtm++ just a couple of minor comments

api/client/batch_client.py
itsjohnward: `json_decode` was always being called, but I moved it up from `get_data_points()` to here since the `raise gen.Return(data.body)` was breaking on 204 No Content responses since they don't have a body.

lib.py equivalent is `resp.json()` here https://github.com/gro-intelligence/api-client/blob/fc39206bf5c81fe63721d9bcc834899add6824a0/api/client/lib.py#L555 but tornado's body is bytes instead of a string, so it doesn't have that method. Have to decode and parse json, like `json.loads(response.body.decode("utf-8"))` to convert it to a python object. Tornado provides `json_decode` that just does those two things in one function.

api/client/batch_client.py
itsjohnward: Yeah, retry_count was more like an num_attempts_count. Updated here for Batch and in https://github.com/gro-intelligence/api-client/pull/175/files for lib as well

nemozen (approved):

PR #178, opened 2020-03-09
handle case where data cannot be normalized return absolute weights

nemozen:
Avoids a division by zero failure when the total is 0.

We do not enforce any assumption on the given series, but in basic
cases where the weighting series is non-negative, total=0 happens when
the data is all 0 or missing so returning 0 weights is most convenient
in applications where there are multiple sets of weights.

### DIFF
--- a/api/client/crop_model.py
+++ b/api/client/crop_model.py
@@ -67,7 +67,7 @@ class CropModel(GroClient):
         self.get_logger().debug('Means = {}'.format(
             list(zip([region['name'] for region in regions], means))))
         # Normalize into weights
-        total = math.fsum([x for x in means if not math.isnan(x)])
+        total = numpy.nansum(means)
         if not numpy.isclose(total, 0.0):
             return [float(mean)/total for mean in means]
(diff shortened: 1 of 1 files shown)


itsjohnward (approved):

api/client/crop_model.py
```diff
@@ -63,11 +64,16 @@ def mapper(region):
                       (df['metric_id'] == entities['metric_id']) &
                       (df['region_id'] == region['id'])]['value'].mean(skipna=True)
         means = list(map(mapper, regions))
-        self._logger.debug('Means = {}'.format(
+        self.get_logger().debug('Means = {}'.format(
             list(zip([region['name'] for region in regions], means))))
         # Normalize into weights
         total = math.fsum([x for x in means if not math.isnan(x)])
```
tolothy: a minor improvement here since we are using numpy, this is equivalent to `numpy.nansum(mean)`

nemozen: done

PR #179, opened 2020-03-11
WEB-679: fix link to understanding data responses page

seungatGro:
image

relative path to html generated by sphinx

### DIFF
--- a/docs/gro-ontology.rst
+++ b/docs/gro-ontology.rst
@@ -21,7 +21,8 @@ The entity types around which the data is organized in Gro are:
 Entity fields
 =============
 Each specific entity will have an `id`, `name`, and various other properties. For example, "Soybeans" is a specific item with `id = 270` and the following information
-::
+
+.. code-block:: python
 
   { 'contains': [1737, 7401, 7578, 7579, 5136, 5021, 5135, 5137, 12418, 5776, 5777, 3035, 3036, 9312, 9284, 6330, 7577, 12728, 12729, 12730, 12731, 12732, 12733, 12734, 12735, 12736, 12737, 10134],
     'definition': 'The seeds and harvested crops of plants belonging to the species <i>Glycine max</i> that are used in the production of oil and both human and livestock consumption.',
@@ -31,7 +32,8 @@ Each specific entity will have an `id`, `name`, and various other properties. Fo
 IDs are unique within a particular entity type, i.e., there's only one item with id = 270.
 
 The above information can be retrieved using the :meth:`api.client.gro_client.GroClient.lookup` function:
-::
+
+.. code-block:: python
 
   client.lookup('items', 270)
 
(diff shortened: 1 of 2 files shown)


seungatGro:
image

fixed reference links mentioned in issue

docs/gro-ontology.rst
```diff
@@ -35,7 +35,7 @@ The above information can be retrieved using the :meth:`api.client.gro_client.Gr
 
   client.lookup('items', 270)
 
-`See this Getting Started section for help on finding entities by name or by id. <understanding-data-responses#look-up>`
+See this Getting Started section for help on finding entities by name or by id. `Understanding Data Responses#lookup <understanding-data-responses#lookup>`_
```
JGrinst: I think we want this to read more like:

`See this Getting Started section for help on finding entities by name or by id.<understanding-data-responses#lookup>`

<understanding-data-responses#lookup> is not text that needs to be seen by the user

docs/gro-ontology.rst
```diff
@@ -52,7 +52,7 @@ And, in the opposite direction, the following will return a list of regions "Cal
 
   client.lookup_belongs('regions', 13055)
 
-`See this Getting Started section for navigation of relationships between entities. <./searching-data.html#lookup-contains>`
+See this Getting Started section for navigation of relationships between entities. `Searching Data#lookup-contains <searching-data#lookup-contains>`_
```
JGrinst: Similar to above comment:

`See this Getting Started section for navigation of relationships between entities.<searching-data#lookup-contains>`

The user doesn't need to see the Searching Data#lookup-contains  text

docs/gro-ontology.rst
seungatGro: image
this is what is generated in html

docs/gro-ontology.rst
seungatGro: image

docs/gro-ontology.rst
JGrinst: Right, but I don't think it should look like that. "Understanding Data Responses#lookup" doesn't look right. It should be more along the lines of what it used to look like:

image

seungatGro: gotcha fixed to updated link style:
image

JGrinst (approved):

docs/querying-data.rst
```diff
@@ -11,15 +11,15 @@ All of the examples in this page refer to a client object, which can
 be initialized as follows:
 ::
```
itsjohnward: Can still provide the language hint, like `.. code-block:: python` instead of the generic `::`

https://www.sphinx-doc.org/en/1.5/markup/code.html#directive-code-block

itsjohnward (changes requested):

docs/querying-data.rst
seungatGro: it is python by default
`The valid values for the highlighting language are:

none (no highlighting)
python (the default when highlight_language isn’t set)`

should specify if otherwise

seungatGro: pushed change to update each caller to epxlicit python blocks

itsjohnward (approved):

PR #182, opened 2020-03-13
in compute_crop_weighted_series use add_data_series() …

nemozen:
Update CropModel to take advantage of improved data series search.

- In compute_crop_weighted_series use add_data_series() to find and add the data series to weight. This performs a real search on series and adds the top result, whereas the old way was searching for the top matching item and metric separately and assuming the pair was the best. 

- Same thing in compute_weights() when finding the weighting series.

Also cleaned up  unused constants, and reduced some logging verbosity.


eliaschneider (approved):
LGTM, love that add_data_series() returns the_data_series variable

PR #185, opened 2020-03-13
Time and frequency ranking 

nemozen:
Add support for time and frequency in Client, and use it in find_data_series()


api/client/gro_client.py
```diff
@@ -280,9 +280,9 @@ def find_data_series(self, **kwargs):
 
         See https://developers.gro-intelligence.com/data-series-definition.html
 
-        This method uses :meth:`~.search` to find entities by name and :meth:`~.get_data_series` to
-        find available data series for all possible combinations of the entities, and
-        :meth:`~.rank_series_by_source`.
+        This method uses :meth:`~.search`, :meth:`~.get_data_series`,
+        :meth:`~.get_available_timeandfrequency` and  :meth:`~.rank_series_by_source`.
```
muzigao: function name should be `get_available_timefrequency` ?

api/client/gro_client.py
```diff
@@ -323,22 +323,36 @@ def find_data_series(self, **kwargs):
...
-            for data_series in data_series_list:
-                if kwargs.get('start_date'):
-                    data_series['start_date'] = kwargs['start_date']
-                if kwargs.get('end_date'):
-                    data_series['end_date'] = kwargs['end_date']
-            all_data_series += data_series_list
-        self._logger.warning("Found {} distinct data series total for {}".format(
-            len(all_data_series), kwargs))
-        for data_series in self.rank_series_by_source(all_data_series):
+            for data_series in self.get_data_series(**entities):
```
muzigao: i noticed that it will be very slow if any of the search_result actually expanded to a list of data series. for example, 
```
next(client.find_data_series(item="Rainfall",
                             metric="Precipitation Quantity",
                             region="United States of America"))
```
one of the  possible combination `({'id': 10081}, {'id': 2100031}, {'id': 1215})` will result in 50000 regions 
should we batch freq ranking for those expanded series (per each search result) ?
also otherwise, in next step,  `rank_series_by_source`  would be comparing expanded series with "at level"  search result?

api/client/lib.py
```diff
@@ -447,6 +447,21 @@ def rank_series_by_source(access_token, api_host, series_list):
             yield series_with_source
 
 
+def get_available_timefrequency(access_token, api_host, **series):
+    def make_key(key):
```
muzigao: would we make this function shared between `get_available_timefrequency` and `get_source_ranking` ?

api/client/__init__.py
```diff
@@ -328,3 +329,35 @@ def get_descendant_regions(self, region_id, descendant_level=None,
         """
         return lib.get_descendant_regions(self.access_token, self.api_host, region_id,
                                           descendant_level, include_historical, include_details)
+
+
+    def get_available_timefrequency(self, **selection):
```
muzigao, line 334: tested a few examples , `get_available_timefrequency`  works as desired.

api/client/__init__.py
```diff
@@ -328,3 +329,35 @@ def get_descendant_regions(self, region_id, descendant_level=None,
...
+        -------
+        list of dicts
+
+            Example::
+
+                 [{
+                    'startDate': '2000-02-18T00:00:00.000Z',
+                    'frequencyId': 3,
+                     'endDate': '2020-03-12T00:00:00.000Z',
+                     'name': '8-day'
```
muzigao, line 355: I noticed that we have an extra "period" field returned from API, could be removed though, viz is not using that neither

api/client/gro_client.py
nemozen: yes get_data_series() was originally  a generator (i.e. yields rather than return). however recent changes made it return a list intead of a generator...... I've put a limit of for now in cfg

api/client/lib.py
nemozen: done

nemozen:
tested and update ptal

muzigao (approved):
retested - function looks good to me! 

(not sure if we'd like to specify that MAX_SERIES_PER_COMB somewhere in the document, but could be annoying if would like to change it in the future, maybe could be user input in the future)

api/client/cfg.py
```diff
@@ -6,3 +6,4 @@
 MAX_QUERIES_PER_SECOND = 10
 TIMEOUT=6000
 MAX_RESULT_COMBINATION_DEPTH=3
+MAX_SERIES_PER_COMB=1000
```
itsjohnward, line 9: With this as 1000, I was finding that "production", "corn", "united states" was taking ~5 minutes to run. Killed it prematurely a few times thinking it was unresponsive before I added some logs and realized it was just looping through thousands of results

Reduced this number down to 5, and the same query only takes 5 seconds. And I got the same top result

It's a trade-off between hit ratio and responsiveness, but 1000 seems really high to me

api/client/gro_client.py
```diff
@@ -370,22 +370,37 @@ def find_data_series(self, **kwargs):
...
-                    data_series['end_date'] = kwargs['end_date']
-            all_data_series += data_series_list
-        self._logger.info("Found {} distinct data series total for {}".format(
-            len(all_data_series), kwargs))
-        for data_series in self.rank_series_by_source(all_data_series):
+            for data_series in self.get_data_series(**entities)[:cfg.MAX_SERIES_PER_COMB]:
+                count += 1
+                self._logger.debug("Data series: {}".format(data_series))
+                # time range affects ranking
+                data_series.pop('start_date', None)
```
itsjohnward, line 381: Doing the inverse seems like it'd be safer here:

```py
data_series = {key: data_series[key] for key in data_series.keys()
               if key in ['metric_id', 'item_id', 'region_id', 'partner_region_id']}
```

just in case we added something to the data_series objects in the future that might break the `frozenset` hashing (like how adding metadata and it being an object messed up pandas hashing before).

api/client/gro_client.py
```diff
@@ -370,22 +370,37 @@ def find_data_series(self, **kwargs):
...
+                if series_hash not in ranking_groups:
+                    ranking_groups.add(series_hash)
+                    if kwargs.get('start_date'):
+                        data_series['start_date'] = kwargs['start_date']
+                    if kwargs.get('end_date'):
+                        data_series['end_date'] = kwargs['end_date']
+                    for tf in self.get_available_timefrequency(**data_series):
+                        ds = dict(data_series)
+                        ds['frequency_id'] = tf['frequency_id']
+                        freq_ranking.append(ds)
```
itsjohnward: I don't think we need this `freq_ranking` list at all.

If instead of `freq_ranking.append(ds)` you just do

```py
for data_series in self.rank_series_by_source([ds])
  yield data_series
```

you'd get the same results since in `rank_series_by_source` it's just going through the input list in order and only ranking sources for one input series at a time. Not re-ordering all the results based on source ranking.

That way, the generator can start yielding results before going through all of the combinations first. That would address my above complaint about `MAX_SERIES_PER_COMB` being 1000 taking a long time.

itsjohnward (changes requested):

api/client/gro_client.py
nemozen: yes  i was planning on doing that in the first place but then decided to  log the intermediate results before source ranking...

nemozen: There's actually a difference in theory though not in practice. We are ranking all series by time and frequency and then re-ranking  by source. So the source ranking could reshuffle the frequencies. E.g. timefreq gives

(freq 1, region1), (freq2, region2) ... 

and then source ranking gives:

(freq2, region2, source1),  (freq1, region1, source2),  

the current implementation of rank_series_by_source does not do that, but we might want that.

nemozen:
> retested - function looks good to me!
> 
> (not sure if we'd like to specify that MAX_SERIES_PER_COMB somewhere in the document, but could be annoying if would like to change it in the future, maybe could be user input in the future)

Right we don't really want ordinary users to change this variable, since they would jut want it to be infinite i.e. look through everything.  Like the COMBINATION_DEPTH,  it should only be used to fine tune  based on back-end search performance

api/client/gro_client.py
nemozen: changed. right now I think the simplication is worth it.

api/client/cfg.py
nemozen: i've been using a lot of examples (6 or 7 crops, 3 countries, districts and provinces).. 100 is too small (especially when you have  e.g a metric= land cover, item = a crop that has a lot of varieties, region = a region with many descendants that have different items, e.g. millet. 
US corn being one of the biggest makes sense, obviously there are  quite a few  items and many many region levels to consider.

 its a trade-off but the purpose of find_data_series is deep exploration so the tradeoff is very biased for quality than speed. the performance sensitive operations are getting data.  
in terms of this PR scope, let's stick to going rfom unlimited to 1000.

api/client/gro_client.py
nemozen: yeah I was hesitating about this but concluded it is more understandable this way. What we're doing is not saying we only want item/metric/region/partner_region  or just all series selection entities like ENTITY_KEY_TO_TYPE above. We are saying we specifically want to remove the things we're going to rank on i.e. time and freq, then source. so I decided to go with explicit removal than explicit addition. 

THere are more comments in the subsequent pr
https://github.com/gro-intelligence/api-client/pull/192/files#diff-603e79178b8e522d0e366a9fca477b24R388

api/client/cfg.py
nemozen: addressed with generator impl below

itsjohnward (approved):

PR #186, opened 2020-03-14
update gro cli to use new search functionality and cleanup 

nemozen:
- use find_data_series in cmd line tool
- delete obsolete constants
- don't set default log level from cfg.py : user can control it directly if they want to change the default


api/client/gro_client.py
```diff
@@ -634,23 +634,12 @@ def main():
...
-        selected_entities['item_id'] = client.search_for_entity('items', args.item)
-    if args.metric:
-        selected_entities['metric_id'] = client.search_for_entity('metrics', args.metric)
-    if args.region:
-        selected_entities['region_id'] = client.search_for_entity('regions', args.region)
-    if args.partner_region:
-        selected_entities['partner_region_id'] = client.search_for_entity('regions',
-                                                                          args.partner_region)
-    if not selected_entities:
-        selected_entities = client.pick_random_entities()
```
itsjohnward, line 658: By not picking random entities, we still get a "random" result if no arguments are given, but it's just the first result that get_data_series returns which is consistent over time. So if I run this without arguments a bunch of times back to back I'll just keep getting

```py
{'metric_id': 15741072, 'metric_name': 'Density', 'item_id': 9125, 'item_name': 'Bulk density (fine earth) at 60 cm deep', 'region_id': 10776, 'region_name': "P'yongan-bukto", 'partner_region_id': 0, 'partner_region_name': 'World', 'frequency_id': 15, 'source_id': 89}
```

A bit less fun than the actually picking a random item each time

itsjohnward (approved):

api/client/gro_client.py
nemozen: fun restored

PR #189, opened 2020-03-19
add missing command

JGrinst:
The command must have got removed in the redesign. Just need to add it back.

### DIFF
--- a/docs/prerequisites.rst
+++ b/docs/prerequisites.rst
@@ -21,7 +21,7 @@ The Gro API Client package is supported both with or without Anaconda. However, 
 Anaconda
 --------
 
-1. Download Anaconda with Python 3.5 or above from `anaconda.com <https://mail.python.org/pipermail/python-dev/2018-March/152348.html>`_ soon approaching, it is recommended you start with Python 3.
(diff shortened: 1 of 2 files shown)


docs/prerequisites.rst
```diff
@@ -21,7 +21,7 @@ The Gro API Client package is supported both with or without Anaconda. However,
 Anaconda
 --------
 
-1. Download Anaconda with Python 3.5 or above from `anaconda.com <https://mail.python.org/pipermail/python-dev/2018-March/152348.html>`_ soon approaching, it is recommended you start with Python 3.
+1. Download Anaconda with Python 3.5 or above from `anaconda.com <https://www.anaconda.com/distribution/>`_ Support for Python 2.7.13 or above is also maintained, but with `its End of Life <https://mail.python.org/pipermail/python-dev/2018-March/152348.html>`_ soon approaching, it is recommended you start with Python 3.
```
itsjohnward, line 24: The EOL's actually passed now (almost 3 months ago at this point), so we could update this language to say something like "As of this writing, the Gro client library still supports Python 2.7.13+, but as Python 2 is now officially past its End of Life, we strongly recommend you use Python 3."

Let's make that its own PR after this though.

itsjohnward (approved):

PR #190, opened 2020-03-23
CLEWS-26398-analogous_years_stop_tsfresh_multiprocessing

joysaikat:
Resolving issues related to celery and multiprocessing (tsfresh). Refer to the CLEWS number and https://github.com/blue-yonder/tsfresh/issues/458 for detail.


api/client/samples/analogous_years/lib/get_transform_data.py
```diff
@@ -26,6 +26,8 @@ def get_data(client, metric_id, item_id, region_id, source_id, frequency_id, sta
                    'source_id': source_id,
                    'frequency_id': frequency_id,
                    'start_date': start_date}
+    client._data_series_list = []
+    client._data_frame = pd.DataFrame()
```
nemozen: hmm this is dangerous...  and could opens up a concurrency can of worms. instead of passing around a client and resetting, just make it a local variable and  create a new one? the only purpose of keeping a client object instance is to keep the data frame and state.

api/client/samples/analogous_years/run_analogous_years.py
```diff
@@ -111,6 +111,8 @@ def main():
     parser.add_argument('--all_ranks', action='store_true', help='Lets you see all the ranks'
                                                                  'as opposed to separate method'
                                                                  'ranks')
+    parser.add_argument('--num_jobs', type=int, default=0,
```
nemozen, line 114: shouldn't the default be 1? you cant run on zero threads ;)

joysaikat: https://github.com/blue-yonder/tsfresh/issues/458

> Sounds like you are not allowed to spawn 'grand-children' from your Pool processes. The extract_features() method spawns two children by default (refer to the API documentation). Try passing n_jobs=0 as an argument to the extract_features() method.

api/client/samples/analogous_years/lib/get_transform_data.py
joysaikat: client is passed as an argument; so should I do `client_1 = client` and then work with `client_1`?

api/client/samples/analogous_years/run_analogous_years.py
joysaikat: Also, celery no longer complaining and sending out emails.

api/client/samples/analogous_years/lib/get_transform_data.py
joysaikat: Discussed offline. Will not client as an argument here.

@nemozen , question should that be part of a different PR. Since, the main problem that I am trying to solve here is not related to this one?

nemozen: ok

nemozen (approved):

api/client/samples/analogous_years/run_analogous_years.py
tolothy: just double checked the doc for tsfresh https://tsfresh.readthedocs.io/en/latest/api/tsfresh.feature_extraction.html#tsfresh.feature_extraction.extraction.extract_features
`n_jobs (int) – The number of processes to use for parallelization. If zero, no parallelization is used.` So we should be good with 0.

PR #191, opened 2020-03-23
use next() for py3

nemozen:


itsjohnward (approved):
Ran in 2.7.16 and 3.8.1 just to be sure. Verified this works 👍 

Random input I tested:

```py
>>> client.growing_degree_days('virginia', 70, '2010-01-01', '2010-12-31', 0.14, 100)
```

PR #192, opened 2020-03-23
CLEWS-26419: add optional result_filter in find_data_series

nemozen:
Passing in a result_filter allows the user to apply any entity  filter in find_data_series (e.g. only a given source, or frequency, or specify region and let item and metric be searched). Compared to users just filtering after getting all the result this is an optimization of speed but also  quality since early filtering allows it to go deeper in the worst cases.

- add optional result_filter in find_data_series
- cleanup implementation of combinatorial search
- use result_filter in CropModel's crop_weighted_series


api/client/samples/crop_models/requirements.txt
```diff
@@ -4,7 +4,6 @@ ipython
 jupyter
 matplotlib
 numpy
-pandas
```
eliaschneider, line 7: why removing pandas?

eliaschneider (approved):
LGTM, only one minor comment

api/client/samples/crop_models/requirements.txt
nemozen: redudant with
https://github.com/gro-intelligence/api-client/blob/development/requirements.txt#L6

PR #193, opened 2020-03-25
CLEWS-26479 fix ranks on repeated analogous year run

joysaikat:
1. Refactoring the code so that the client is re-initiated before every run of analogous years. Failing to do so may give rise to undesirable outcome in terms of the ranks output.

Note: After merging this `gro/experimental/alerts/gro_alert_task.py`, `gro/experimental/alerts/gro_alert_services.py` and `gro/experimental/alerts/run_gro_alert.py` may need change.


api/client/samples/analogous_years/lib/final_ranks_computation.py
```diff
@@ -170,14 +172,14 @@ def combined_methods_distances(dictionary_of_df):
     return rank_df
 
 
-def analogous_years(client, data_series_list, initial_date, final_date,
+def analogous_years(api_token, data_series_list, initial_date, final_date,
```
tolothy, line 175: We should think about a way to make it backward compatible, otherwise it will break any existing client side code. Maybe we could make the first argument a token_or_client_instance? thoughts?

tolothy: same for all remaining functions that the argument has changed, but this is the most important case.

joysaikat: I have an idea (very wasteful).
`def analogous_years(client, data_series_list, initial_date, final_date,`...
`client = GroClient(client.api_host, client.access_token)`

joysaikat: Discussed offline with John and Yiqing. Using a different hack to resolve the issue for now; so that things do not fall apart in `email_alert`.

api/client/samples/analogous_years/run_analogous_years.py
```diff
@@ -52,14 +55,15 @@ def check_if_exists(entity_type, entity_value, client):
 
 
 def get_data_series_list(region_id_list, item_id_list, metric_id_list, source_id_list,
-                         frequency_id_list, client):
+                         frequency_id_list, api_token):
     # checking if the length of the list for metric_id, item_id, source_id and
     # frequency_id match
+    client = GroClient(API_HOST, api_token)
```
tolothy: I think this line is redundant now?

api/client/samples/analogous_years/run_analogous_years.py
```diff
@@ -114,29 +117,29 @@ def main():
     parser.add_argument('--num_jobs', type=int, default=0,
                         help='number of parallel processes for tsfresh')
     args = parser.parse_args()
-    client = GroClient(API_HOST, args.groapi_token)
+    # client = GroClient(API_HOST, args.groapi_token)
```
tolothy: delete

api/client/samples/analogous_years/run_analogous_years.py
joysaikat: Addressed! Thanks!

api/client/samples/analogous_years/run_analogous_years.py
joysaikat: Complied!

api/client/samples/analogous_years/lib/get_transform_data.py
```diff
@@ -28,6 +28,8 @@ def get_data(client, metric_id, item_id, region_id, source_id, frequency_id, sta
                    'start_date': start_date}
     client.add_single_data_series(data_series)
     data = client.get_df()
+    # TODO: select data from dataframe such that the date start_date restrictions are maintained
+    #  and the client does not need to be re-initiated in a single jupyter notebook
```
nemozen: why not just  take `df.loc[df.end_date >= start_date][['end_date', 'value']]`  ?

it seems strange that you want to fill in the beginning of the series with a duplicate of the first point ... either the data is there and you don't need to fill it or it's missing and you should fill in nulls not assume a constant right?

nemozen (commented):
the change looks good just one question

api/client/samples/analogous_years/lib/get_transform_data.py
joysaikat: 1. The answer to the first question - The reason I did not take this route initially is - I thought I am not changing the data type of the variable `start_date` to datetime until later, which is not the case. So, should I revert back all the changes and just incorporate this?

2. I think this is separate from your first question. If I understand you correctly your concern is about the following lines 
 ```
 if data['end_date'].iloc[0] > start_date:
        new_value = data['value'].iloc[0]
        new_row = pd.DataFrame({'end_date': [start_date], 'value': [new_value]})
        data = pd.concat([new_row, data[:]]).reset_index(drop=True)
```
I had to put it there so that there is a common start date for all the series. Otherwise, the function `loop_start_date` was not producing desired result. Example: If I add rainfall along with ENSO, I was not getting the correct `loop_start_date`.

Comment (joysaikat) [reply api/client/samples/analogous_years/lib/get_transform_data.py]: Discussed with Yiqing. Keeping the previously done changes; but also incorporating the code Nemo suggested. The idea is passing `access_token` for the client instead of `client` may be a better idea.

Comment (nemozen) [reply api/client/samples/analogous_years/lib/get_transform_data.py]: re: 1, thx looks good.

re: 2... yes that was separate point. I understand you are padding it to make it line up with start date. My concern is 
a)  if the first data point p is such that  p.start_date <= start_date  < p.end_date, thenthe partially covered p.value is overweighted instead of under.  For example if it's a weekly series and the data points start,end and value are like:
1/11, 1/17, v1
1/18,1/24,v2
and now you get a start_date of 1/20. This would return 
1/20,v2
1/24,v2
This is imprecise as, the true value for the period ending 1/20 is closer to v1 than v2.

b) More importantly if the data starts later, i.e.  start_date < p.start_date <= p.end_date, say start_date is 1/5 in the above example, then you are returning
1/5,v1
1/11,v1
but it would be more accurate to retrun
1/5,null
1/11,v1

To fix both you could change the downstream logic to deal with data points as representing a value over an interval of time ((start_date, end_date), value) and deal with boundaries as needed. But that would eb a bigger change than this PR.  

For now, i think you shuold at least fix scenario b).  e.g. also keep start_date column and 
```
if data['end_date'].iloc[0] > start_date:
    if data['start_date'].iloc[0] <= start_date:
          new_value = null
   else:
          ....   
```

Comment (joysaikat) [reply api/client/samples/analogous_years/lib/get_transform_data.py]: Your comments are spot on in terms of initial point being unfairly weighed. 

There is at least one other situation where the first point is weighed more. The case when the data series is not sorted by date. John assured me that does not happen with Gro data series. A situation like that would arise if I stuck with this code and fixed the failing unit_test. We did consider this and arrived at the decision that 1 data point weighed against 365 data point would not be enough to change the rank monumentally. Hence, I kept this code. 

Obviously, there are cases where I would be completely wrong - For example the worst scenario in a data series with half yearly frequency. or even monthly frequency. That is why there is a disclaimer that this works best for daily frequency data series.

Comment (joysaikat) [reply api/client/samples/analogous_years/lib/get_transform_data.py]: > but it would be more accurate to retrun
> 1/5,null
> 1/11,v1

The way it is set up, analogous_years need to have all the data series at the initial and final date with non-null value (computing euclidean distance with null value would not be possible). So, I do not think null would work.

More importantly, in a situation where the start_date for the series is later than user provided initial date. The application will not run. Additionally, the function `loop_start_date` determines which year and day should be the starting date for the overall computation. That date is then applied to each series.

Comment (joysaikat) [reply api/client/samples/analogous_years/lib/get_transform_data.py]: Discussed offline: there needs to be a separate PR to make sure that I do not drop the series start_date values. As they are valuable information in case of monthly, semi-annual frequency.

Comment (joysaikat) [reply api/client/samples/analogous_years/lib/get_transform_data.py]: I stuck with the previously discussed done in this PR, with a TODO for the next PR. I look towards almighty reviewers for approval!!

**Review by nemozen — APPROVED** [review]

File: api/client/samples/analogous_years/lib/get_transform_data.py [hunk 5]
```diff
@@ -28,8 +28,10 @@ def get_data(client, metric_id, item_id, region_id, source_id, frequency_id, sta
                    'start_date': start_date}
     client.add_single_data_series(data_series)
     data = client.get_df()
```
nemozen, line 30: here you can use the new functionality to only get one series:

`data = client.get_df(index_by_series=True).loc[[tuple(metric_id,item_id, region_id, 0, source_id, frequency_id)], :]
`

nemozen (commented):
one more thing.. now that we have the index_by_series you can improve get_data(). doesn't have to  be in this PR, but please add TODO

api/client/samples/analogous_years/lib/get_transform_data.py
joysaikat: I think the documentation about this function here need more detail https://developers.gro-intelligence.com/api.html?highlight=index_by_series .

I am not sure what performance improvement would I see due to indexing by series, though?

nemozen: It's simply giving a view on the df that only accesses the rows of the particular series and excluding all other rows. In this case since you've just modified to only have one data series I guess it makes no difference but it would have solved the original issue :)

joysaikat: Okay, thanks for explaining. The code that worked for me was `client.get_df(index_by_series=True).loc[(metric_id, item_id, region_id, 0, source_id, frequency_id), :]`
More importantly, this will not work in a situation where the region breaks down into subregions. Example: corn belt. May be removing `region_id` and then selecting the rows might work.

PR #194, opened 2020-03-29
CLEWS-15676: add option to index by series in get_df, use it in GDH 

nemozen:
this index gives users a shortcut to retrieve one series at a time from within the df

Screenshot from 2020-03-30 14-31-31

Screenshot from 2020-03-30 14-31-58

Screenshot from 2020-03-30 14-32-16

### DIFF
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -235,8 +235,8 @@ class GroClient(Client):
     def GDH(self, gdh_selection, **optional_selections):
         """Wrapper for :meth:`~.get_data_points`. with alternative input and output style.
 
-        The selection of data series to retrieve is encoded in a
-        'gdh_seletion' string of the form
+        The data series selection to retrieve is encoded in a 
+        'gdh_selection' string of the form
         <metric_id>-<item_id>-<region_id>-<partner_region_id>-<source_id>-<frequency_id>
 
         For example, client.GDH("860032-274-1231-0-14-9") will get the
@@ -255,7 +255,7 @@ class GroClient(Client):
         ------
         pandas.DataFrame
 
-            the main DataFrame :meth:`~.get_df`. with the :meth:`~.get_data_points`. results for requested series.
+            the subset of the main DataFrame :meth:`~.get_df`. with the requested series.
 
         """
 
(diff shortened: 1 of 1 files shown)


api/client/gro_client.py
```diff
@@ -257,19 +259,16 @@ def GDH(self, gdh_selection, **optional_selections):
...
         # add optional pararms to selection
         for key, value in list(optional_selections.items()):
-            if key not in entity_keys:
+            if key not in DATA_SERIES_UNIQUE_COLS:
                 selection[key] = value
 
         self.add_single_data_series(selection)
-        df = self.get_df()
-        return df
+        return self.get_df(index_by_series=True).loc[[tuple(entity_ids)], :]
```
muzigao: this `.loc[[tuple(entity_ids)], :]` part will throw an error if given selection doesn't have at-level data. e.g client.GDH("860032-10009-1215-0-14-9")

api/client/gro_client.py
```diff
@@ -64,13 +63,18 @@ def get_df(self, show_revisions=False):
...
-
+            If index_by_series is set, the dataframe is indexed by series.
+            See https://developers.gro-intelligence.com/data-series-definition.html
         """
         while self._data_series_queue:
             data_series = self._data_series_queue.pop()
             if show_revisions:
                 data_series['show_revisions'] = True
             self.add_points_to_df(None, data_series, self.get_data_points(**data_series))
+        if index_by_series:
```
muzigao, line 74: tested with a couple of examples, data frame multi-index  part looks good to me !

api/client/gro_client.py
nemozen: that's the intended behavior... gdh is expected to return only that series exactly.

muzigao: yeah I understand that - but it's showing 

`KeyError: "None of [MultiIndex([(860032, 10009, 1215, 0, 14, 9)],\n           names=['metric_id', 'item_id', 'region_id', 'partner_region_id', 'source_id', 'frequency_id'])] are in the [index]"`

should we return an empty data frame or something ?

stephenbernardi (approved):
LGTM. In the future, can we add an improvement to convert index ids to names?

api/client/gro_client.py

muzigao (approved):

PR #195, opened 2020-03-30
CLEWS-17178: optimize GroClient data retrieval for dataframe

nemozen:
When a user  adds the same series repeatedly currently we just do it naively i.e. don’t try to deduplicate.  This is inefficient in scenarios where the user is rerunning the same functions many times (e.g. developing a model iteratively in a notebook for example). At the same time, in most use cases, the underlying data does not change within a single run of a model so we don't need  incremental updates (though that might be useful in the future in conjuction with load/save to disk for longer term data state).

This change deduplicates and only retreives each series once.

### DIFF
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -42,7 +42,7 @@ class GroClient(Client):
     def __init__(self, api_host, access_token):
         super(GroClient, self).__init__(api_host, access_token)
         self._logger = lib.get_default_logger()
-        self._data_series_list = []  # all that have been added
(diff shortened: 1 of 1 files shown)


eliaschneider (approved):
LGTM

tolothy (approved):
Tested with 
```
client.add_data_series(
	metric='cash price', item='sunflower oil', region='china')
print(client._data_series_list)
client.add_data_series(
	metric='cash price', item='sunflower oil', region='china')
print(client._data_series_list)
```
before and after this change. After the change only one data series is printed twice.

PR #196, opened 2020-04-02
CLEWS-26695 Top N method

itsjohnward:
I added several usage examples to the `Client.get_top()` docstring, highlighting:

1. the minimal case: full time range, default n
2. top in a single year
3. a specific time range, with n>5

PTAL at the docstring in particular for documentation clarity.


cn1036 (approved):
tested locally. new method looks good to me!

PR #198, opened 2020-04-03
CLEWS-17178 Add index names to get_df

itsjohnward:
Follow-up to https://github.com/gro-intelligence/api-client/pull/194

Now, you can use `df.index.names` to take an index value and re-associate it back to metric_id/item_id/etc.:

```py
df = client.get_df(index_by_series=True)

for series in df.index.unique():
    series_name = ' '.join([client.lookup(ENTITY_KEY_TO_TYPE[type_id], series[i])['name']
                            for i, type_id in enumerate(df.index.names)])
    print(series_name)
```

Before, you would have needed to know which order the series attributes are indexed in.

### DIFF
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -1,6 +1,4 @@
 from __future__ import print_function
-from api.client import cfg, lib, Client
-from api.client.constants import DATA_SERIES_UNIQUE_TYPES_ID, ENTITY_KEY_TO_TYPE
 from builtins import str
 from builtins import zip
 from random import random
@@ -9,8 +7,13 @@ import functools
 import getpass
 import itertools
 import os
-import pandas
 import sys
+
+from api.client import cfg, lib, Client
+from api.client.constants import DATA_SERIES_UNIQUE_TYPES_ID, ENTITY_KEY_TO_TYPE
+from api.client.utils import intersect
+
+import pandas
 import unicodecsv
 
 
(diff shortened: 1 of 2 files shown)


stephenbernardi (approved):
client.add_single_data_series({'metric_id': 8671080, 'item_id': 274, 'region_id': 1215, 'partner_region_id': 0, 'source_id': 63, 'frequency_id': 9})
df = client.get_df()
dict(zip(df.index.names, df.iloc[0].name))
Out[24]: 
{'metric_id': 8671080,
 'item_id': 274,
 'region_id': 1215,
 'partner_region_id': 0,
 'frequency_id': 9,
 'source_id': 63}

api/client/gro_client.py
```diff
@@ -62,9 +62,12 @@ def get_df(self, show_revisions=False, index_by_series=False):
                 data_series['show_revisions'] = True
             self.add_points_to_df(None, data_series, self.get_data_points(**data_series))
         if index_by_series:
-            return self._data_frame.set_index([c for c in filter(
```
nemozen, line 65: hmm filter is more readabe, why change this?

nemozen (changes requested):
change looks good just a minor comment

api/client/gro_client.py
itsjohnward: Controversial opinion, but filter+lambda vs list comprehension debate aside:

In this case it was already using a list comprehension to begin with. So, going from 3 elements (comprehension+filter+lambda) down to 1 (just list comprehension) seems like the less complex version to me

nemozen: yes  should just be `list(filter(...))` rather than hybrid.  computationally both are trivial but readability-wise i've settled on  filter() as the common pattern for this kid of stuff here (i'm sure some inconsistencies  here and there but let's reduce them).

itsjohnward: Okay, changed back to the `list(filter(lambda...))` syntax and moved into a named `intersect` function

nemozen (approved):

PR #199, opened 2020-04-06
CLEWS-26704 add api host as kwarg in ay

joysaikat:
I suspect the error here https://grointelligence.atlassian.net/browse/CLEWS-26704 is due to `api_host=api.gro-intelligence.com` for STAGE. So, I am 
1. adding `api_host` as a keyword argument for future with default value of `api.gro-intelligence.com` in the relevant places.
2. Currently fetching `api_host` from the instance itself.

### DIFF
--- a/api/client/samples/analogous_years/lib/final_ranks_computation.py
+++ b/api/client/samples/analogous_years/lib/final_ranks_computation.py
@@ -19,6 +19,8 @@ from api.client.samples.analogous_years.lib import \
     feature_extractions, \
     get_transform_data
 
+DEFAULT_API_HOST = 'api.gro-intelligence.com'
+
 
 def common_start_date(client, data_series, provided_start_date=None):
     """Computes the earliest available start date from which all gro-data_series have data"""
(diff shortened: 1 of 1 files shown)


api/client/samples/analogous_years/lib/final_ranks_computation.py
```diff
@@ -51,9 +50,10 @@ def enso_data(start_date):
     return enso_data_series
 
 
-def get_file_name(api_token, data_series_list, initial_date, final_date):
+def get_file_name(api_token, data_series_list, initial_date, final_date,
+                  api_host='api.gro-intelligence.com'):
```
tolothy: to accommodate the scenario when the default host has to change again, you can still have a global var, sth like `DEFAULT_API_HOST = 'api.gro-intelligence.com'`, and default `api_host=DEFAULT_API_HOST`, to avoid changing the host address in every single call.

joysaikat: Done

nemozen (approved):

tolothy (approved):
lgtm now

itsjohnward (approved):

PR #202, opened 2020-04-08
update brazil soybeans yield notebook

nemozen:
new version of crop weighting and gro client uses frequency ranking so we don't need to filter out 16-day from 8day. also don't need to load the province ndvi separately just do it within crop weighted is enough


api/client/samples/crop_models/brazil_soybeans.ipynb
```diff
@@ -57,7 +57,7 @@
      "name": "stdout",
```
tolothy, line 55: <p>The chart is conflicting what the description above this cell says? In the description syas <code>the crop-weighted NDVI series was built with weights normalized to 1.0</code>, therefore in the chart the blue line -- crop_weighted_value -- should have a value on y axis between 0 and 1? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/202/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

nemozen: fixed

tolothy: looks good now

tolothy (approved):
lgtm now.

PR #203, opened 2020-04-08
stocks to use notebook: specify frequency id in API call

stephenbernardi:
API was throwing an error because frequency_id was not included


JGrinst (approved):
Worked for me! Looks like we removed the sample outputs, could those be added back? I think that's ideal for when the notebook is viewed from the showcase

itsjohnward (approved):

stephenbernardi:
> Worked for me! Looks like we removed the sample outputs, could those be added back? I think that's ideal for when the notebook is viewed from the showcase

ah, yeah. I am re-running now and will update with the ouput

PR #204, opened 2020-04-09
CLEWS-26704-1 Bug fix after adding api host

joysaikat:
This bug gives me opportunity to ponder if it would be better to pass api_token as a kwarg as well like api_host?


itsjohnward (approved):

tolothy (commented):
lgtm
i think we dont want to pass in token as a kwargs. there is no good default value for that.

joysaikat:
How about the environment variable: `GROAPI_TOKEN`

PR #205, opened 2020-04-09
CLEWS-25139 Client->GroClient

itsjohnward:
1. Move over all the Client methods into GroClient
2. Leave an alias for Client->GroClient so `from api.client import Client` still works


seungatGro (approved):
tested Client backwards compatibility and GroClient use of new fn. LGTM

PR #208, opened 2020-04-10
CLEWS-26731 analogous years resampling function

joysaikat:
1. The plan of this PR is to simplify date logics and make the interpolation functions more reasonable. Follow the discussion listed here: https://github.com/gro-intelligence/api-client/pull/193#discussion_r401093430
2. As an addition some unnecessary code will be deleted which is a remnant from earlier PRs:   `CLEWS-26479-fix_ranks_on_repeated_ay_run` and ` CLEWS-26704-add_api_host_as_kwarg_in_ay`. That is, client is no longer passed as an input to the function `analogous_years` in `api/client/samples/analogous_years/final_ranks_computation.py`. Hence, unnecessary lines will be deleted.

Result:
Ran `python run_analogous_years.py -m 2540047 15531082 -i 3457 7382 -r 100000100 -s 26 43 -f 1 1 --initial_date 2019-03-01 --final_date 2020-02-29 --ENSO --all_ranks`
and generated the following csv file (with the ranks were saved in the folder `ranks_csv`)
Screen Shot 2020-04-21 at 20 31 02


api/client/samples/analogous_years/lib/final_ranks_computation.py
```diff
@@ -22,32 +22,32 @@
 DEFAULT_API_HOST = 'api.gro-intelligence.com'
 
 
-def common_start_date(client, data_series, provided_start_date=None):
```
tolothy: I think it is not necessary to change every occurrence of `start_date` to `start_date_bound`, which is making this pr too long

api/client/samples/analogous_years/lib/final_ranks_computation.py
```diff
@@ -78,17 +78,16 @@ def time_series(client, data_series, initial_date, final_date):
     :return: A dataframe with data from the relevant dates
     """
     logger = client.get_logger()
-    entities = {'metric_id', 'item_id', 'region_id', 'source_id', 'frequency_id', 'start_date'}
```
tolothy, line 81: haha this proved my point earlier. just replacing every occurrence of `start_date` by `start_date_bound` is a potential source of bugs

api/client/samples/analogous_years/lib/final_ranks_computation.py
```diff
@@ -78,17 +78,16 @@ def time_series(client, data_series, initial_date, final_date):
...
+    entities = {'metric_id', 'item_id', 'region_id', 'source_id', 'frequency_id', 'start_date_bound'}
     discard = []
     for entity_type in data_series:
         if entity_type not in entities:
             discard.append(entity_type)
     for entity_type in discard:
         data_series.pop(entity_type)
```
tolothy: the above 6 lines feel a bit wordy and non-pythonic. what about just `data_series = {k: data_series[k] for k in data_series if k in entities}` if that is what you are trying to do here?

api/client/samples/analogous_years/lib/final_ranks_computation.py
```diff
@@ -78,17 +78,16 @@ def time_series(client, data_series, initial_date, final_date):
...
     data = get_transform_data.get_data(client, **data_series)
-    consolidated_data = get_transform_data.combine_subregions(data)
```
tolothy, line 89: what is the reason for moving this call from within `get_data` to out here?

joysaikat: Think of corn belt. It will have ten rows with same date, which will make resampling by nearest method more complex. Hence, I moved it upstream. Now, I do not have to use a fill function separately.

api/client/samples/analogous_years/lib/final_ranks_computation.py
joysaikat: `haha` :-o

joysaikat: You are right, I started this PR with the idea that the organic `start_date` (that comes with the data series itself) should be separated from the _start date bound_, human or the program generated. Because, I wanted to use the organic `start_date`. Otherwise, it could become a readability issue. 

Hence, I replaced all the occurrences of `start_date` which came from the `start_date` trickling down as a bound from the `analogous_years` function call. 

This is the only place that I remember that `start_date` had to be replaced with `start_date_bound`, which I did not foresee. 

After reorganizing and streamlining the logic, I did not have to use the organic `start_date` at all. So, I can potentially reverse all the changes. However, I want to keep the current naming convention in case in the future (or even in this PR) we end up using the organic `start_date`.

api/client/samples/analogous_years/lib/final_ranks_computation.py
joysaikat: Replied to the comment below.

api/client/samples/analogous_years/lib/final_ranks_computation.py
joysaikat: Thanks! Done!

api/client/samples/analogous_years/lib/final_ranks_computation.py
joysaikat: Discussed offline. Reversed some of the naming changes and restricted most of the changes to one module for the sake of brevity.

api/client/samples/analogous_years/lib/get_transform_data.py
```diff
@@ -9,48 +9,35 @@
 import pandas as pd
 
 
-def get_data(client, metric_id, item_id, region_id, source_id, frequency_id, start_date):
+def get_data(client, metric_id, item_id, region_id, source_id, frequency_id, start_date_bound):
     """
```
tolothy, line 13: I think now this function does multiple things, it is good to have a description in the doc string, such as `read data from api into a dataframe that is resampled to daily frequency ...`

joysaikat: Thanks! Done!

api/client/samples/analogous_years/lib/get_transform_data_test.py
```diff
@@ -37,15 +38,13 @@ def create_test_data_datetime():
...
-    expected = expected[['end_date', 'value']]
-    new_value = expected['value'].iloc[0]
-    new_row = pd.DataFrame({'end_date': [start_date], 'value': [new_value]})
-    expected = pd.concat([new_row, expected[:]]).reset_index(drop=True)
     expected.loc[:, 'end_date'] = pd.to_datetime(expected['end_date'])
     test_data = get_transform_data.get_data(client, 'metric_id', 'item_id', 'region_id',
-                                            'source_id', 'frequency_id', start_date)
+                                            'source_id', 'frequency_id', start_date_bound)
+    expected = get_transform_data.combine_subregions(expected)[['end_date', 'value']]
+    expected = expected.resample('D').nearest()
```
tolothy: for unit testing purpose, expected results should be given explicitly, instead of being computed step by step according to the function logic

tolothy:
@joysaikat please also comment on what test case you have run and the results

api/client/samples/analogous_years/lib/get_transform_data_test.py
joysaikat: The idea here is I am not trying to unittest the `resample` function here. So, I used the `resample` function and `combine_subregions` has already been tested below.

tolothy: and no testing `combine_subregions`. i get what you mean but what a unit test for is to make sure the function as a whole, not part of the logic, is doing what you would expect, by comparing an explicit expected result with a test run result.

joysaikat: The `combine_subregions` has been tested later. So, I did not test it separately. To your point, I do not think there is a real need to unit test this function. Because, I `combine_subregions` is tested below; `resample` must have been tested elsewhere. So, is commenting out this the more practical approach?

tolothy: i still see a purpose of testing this function as how it calls different functions, and in which order. as a unit test does not care about how a function is implemented, it only cares what is a test case input and what the expected output is in that case.

api/client/samples/analogous_years/lib/final_ranks_computation.py
nemozen: yeah not sure this is helping. what do you mean by start_date_bound?  we use (start_date,end_date) for points, series, and user selected time intervals.

api/client/samples/analogous_years/lib/final_ranks_computation.py
```diff
@@ -47,7 +47,7 @@ def enso_data(start_date):
                         'item_id': 13495,
                         'region_id': 0,
                         'source_id': 124,
-                        'start_date': start_date,
+                        'start_date_bound': start_date,
```
nemozen: definitely unhelpful to change thi svariable name :)    this object is now incompatible with the thing we call "data_series" everywhere.  treat "data_series" as  a  subclass of "dict" ,, i.e. the fields are properties with standard meaning

joysaikat:
> @joysaikat please also comment on what test case you have run and the results
I have now done that @tolothy

api/client/samples/analogous_years/lib/final_ranks_computation.py
joysaikat: I wanted to distinguish between `start_date_bound` (a date provided by the user or computed by the program) versus the `start_date` which is part of gro data series for better readability. Hence, I use the term in the code.

api/client/samples/analogous_years/lib/final_ranks_computation.py
joysaikat: Okay, I changed the variable `start_date_bound` whenever it appeared as part of a dictionary of gro data selection.

api/client/samples/analogous_years/lib/get_transform_data_test.py
joysaikat: I have updated the data and unit test and in doing so fixed a bug :)

nemozen (approved):

PR #209, opened 2020-04-11
Notebooks with cereal (corn, wheat, barley, sorghum and millet) yield models for Ethiopia and Kenya

nemozen:


eliaschneider (approved):
LGTM

PR #214, opened 2020-04-17
CLEWS-27227: memoization on lookups

itsjohnward:
Separate out `lookup_single` and `lookup_batch` from `lookup` so that `lookup_single` can take advantage of memoization.

Before https://github.com/gro-intelligence/api-client/pull/161/files#diff-7ed38827a2c4e58f0dd776f8639dc95cL299 we had memoization on the `lookup()` function so that if you called it many times within the same script, it wouldn't duplicate requests.

It was removed when requesting lists of entities was added as a feature because lists can't be used as hashable inputs to memoization. By splitting up lookup by single id and lookup by list of ids, we can still take advantage of memoization on the former.

Future work: implement memoization such that that only requests the difference between requested entities and already-cached entities

### DIFF
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -242,30 +242,39 @@ def list_available(access_token, api_host, selected_entities):
         raise Exception(resp.text)
 
 
+@memoize(maxsize=None)
+def lookup_single(access_token, api_host, entity_type, entity_id):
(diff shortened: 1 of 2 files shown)


seungatGro (approved):
LGTM. tested with local server to verify requests are not repeated in a process

PR #216, opened 2020-04-26
Link fixes

JGrinst:


itsjohnward (approved):
Diff looks good. Also did a find-all for any `".md"` or `".rst"` in `.md` and `.rst` files, and looks like these were the only ones.

Should also be able to remove `".html"`s, but those do no harm.

PR #217, opened 2020-04-27
update prevent plant notebook with 2020 changes

tolothy:
Main changes:

- kick off 2020 in-season predictions
- source updates and replacements: precipitation is read from GPM instead of TRMM; Two weeks extended precipitation forecast from GFS is an option. 
- some bug fixes to account for a county in the corn belt with no data.


ash-gro:
Looks ok on high level, i.e. assuming various new concat and merge operation does not screw up alignment. I think merge sometimes changes row order (even though it is not supposed to), so I tend to insert sort_index all over the place for safety. These potential alignment issues can't really be  checked without looking at actual dataframes, so you are better positioned to do this.

 I think interpolation I used will extent last loaded values as constants (or maybe even set them to 0) beyond what's loaded (all the way till the end of the year), so you might be able to generate some model projections. But I am not sure you want to use such constant extrapolations them in actual projection computation. Last year all data needed were already in, so it was not an issue, but this year you probably want to ensure that 2020 projection is only generated for model ending week 18 or 19 - whenever last GFS precipitation projection ends. Going beyond that is probably not warranted.

 Finally, I am a bit surprised that INS (2011-2019) accuracy of final GBT models appears to be rather poor (much worse than what we saw last year, I think). It could indeed be that 2019 year being so huge relative to others causes the model to overestimate other years to get closer to the 2019 number, but I don't like that it seems to overestimate even 2019 itself. But if there is no apparent bugs  (such as possible alignment issues I mentioned above), it is what it is, I guess - after all, this is not a project to redo the model (unless higher powers want to make it into one), just to rerun it for 2020.

Finally, on the missing counties for geopandas query - I don't think this was an issue last year. Were new counties introduced sinc ethn?

tolothy:
Update: pushed a fix for `df_all`, which did not have lat/lon filled in properly at the beginning.

@ash-gro regarding your questions:
1. alignment is the exact reason why i switched to `merge` and `concat` in specific cases, namely to create one single df to include both features and target, to make sure features and targets are on one-to-one correspondence.  
2. I was concerned about interpolation at the beginning as well, especially the extrapolation in the end of the time series. As you said we can make sure only models prior to a time period (e.g. models of week 14 through week 16 if not including weather forecast, or through 18 if including weather forecast) are presented to minimize the impact of extrapolations. I can make an automatic cut-off is that's desired.
3. the bug fix above have solved the problem for bad in-sample performance as in the second chart for each period. out-of-sample performances should not change though after the bug fixes. they are even slightly better than last year if you look at the numbers

ash-gro:
<p>Oh, so THAT's what this warning was about.  The dataframe simply did not have lon/lat coordinates in it but, when asked for them, pandas in its misguided wisdom simply filled in with something and gave this obscure warning:</p><p>___________</p><p>/home/yiqing/.local/lib/python3.6/site-packages/pandas/core/indexing.py:979: FutureWarning:</p><p>Passing list-likes to .loc or [] with any missing label will raise</p><p>KeyError in the future, you can use .reindex() as an alternative.</p><p> </p><p>See the documentation here:</p><p>https://pandas.pydata.org/pandas-docs/stable/indexing.html#deprecate-loc-reindex-listlike</p><p>return self._getitem_nested_tuple(tup)</p><p>______________</p><p> </p><p>I just hate these pandas internal warnings without any reference to the user code which caused them - these are next to impossible to trace without removing potentially offending lines one-by-one. Moreover, they are often benign, so natural tendency is to just ignore them if some result is generated ...</p><p> </p><p>And yes, INS results look much good now.</p>

---

 <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/217/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ash-gro:
<p>&gt;&gt;out-of-sample performances should not change though after the bug fixes. they are even slightly better than last year if you look at the numbers</p><p> I assume you refer to the skip-a-year CV-like results. These should indeed be the same as before, but the 2020 full xgb model projection will change, I think.</p>

---

 <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/217/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tolothy:
<p>yes it did. you can see the changes from the diff in the notebook</p>

---

 <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/217/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tolothy:
<p>done on all cases</p>

---

 <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/217/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ash-gro:
On using list of regions instead of getting one-by-one, I think there is a limit on how long the list can be (for example, I don't think one can't get entire world on county level in a single request). But I guess list of counties for a single US state should be ok.

For the code
```
if not in_season_run:
            df_interp.interpolate(inplace=True, limit_direction='both')
        else:
            ....
             df_interp.interpolate(inplace=True, limit_direction='both')
```
wouldn't it be easier to put interpolation outside the if (and have only one if-branch)

Finally, comments (in data loading cells)
```
# pd.interpolate will do linear interpolation ignoring index for multiindexed frames
# which is exactly what we need in this case
```
and
`# internal concat is to add index layer, external one to assemble different frames into one`
is out of date and does not correspond to the code now (interpolated frame is not multiindex and there is only one concat)

itsjohnward:
> On using list of regions instead of getting one-by-one, I think there is a limit on how long the list can be (for example, I don't think one can't get entire world on county level in a single request). But I guess list of counties for a single US state should be ok.

Regarding lists of regions, there is an upper limit: ~300

For districts per US state, you should be okay - the max is Texas with 254.

```py
for state in client.get_descendant_regions(1215, 4):
    print(state['name'], len(client.get_descendant_regions(state['id'], 5)))
```

There's a `api.client.utils.list_chunk` helper function I've been using as a standard way of doing it (it defaults to 50 to be safe, but if you really want to, you can play with the `chunk_size` kwarg). For example:

```py
for region_chunk in list_chunk(region_ids):
    get_data_points(**other_selections, region_id=region_chunk)
```

tolothy:
<p>I really hope we can refactor that once for all when we are ready to expose some reading utility functions from bitbucket such as https://bitbucket.org/grointelligence/gro/src/e236a3377842eb708fd11dd9500a9dc1b9f4441a/xm/prevent_plant/read.py#lines-64, so that we dont have to maintain the two versions separately. </p>

---

 <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/217/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tolothy:
<p>same here. we simplified the logic by a lot in https://bitbucket.org/grointelligence/gro/src/e236a3377842eb708fd11dd9500a9dc1b9f4441a/xm/prevent_plant/read.py#lines-27 by using our API reader. once we expose that part to the client, we can reduce the amount of redundancy here by a lot. Maybe we should just make this a TODO to refactor all the readings once for all after our internal reader is added to the API client as a utility?</p>

---

 <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/217/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tolothy:
<p>Updated. I Sill have them in two separate cells to clearly separate the cases for soil moisture and precipitation, since precip needs special treatment from weather forecast, but the data collection is now through <code>add_single_data_series</code>  and <code>get_df</code></p>

---

 <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/217/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tolothy:
<p>TODO added.</p>

---

 <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/217/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

nemozen (approved):

nemozen:
confirmed with @tolothy that it was tested with py3. shippable tests are passing despite "unstable"

PR #218, opened 2020-04-28
CLEWS-27625 Improve test coverage of GroClient

itsjohnward:
## Goal

cover all documented GroClient methods or major functionality with at least one unit test.

## Parts

1. Add a coverage config file
2. Add mocks for `lib.py` methods. Assume that the `requests.get()`/`get_data()` portion of those methods are covered already by lib_test.py so we don't need to re-test.
3. Add tests using those mocks

## Testing

To run the unit tests locally, make sure testing dependencies are installed

```sh
$ pip install .[test]
```

and run

```sh
$ pytest --cov
```

If you have a coverage viewing tool in your IDE, like https://marketplace.visualstudio.com/items?itemName=brainfit.vscode-coverage-highlighter you can generate xml files for that to use, like so

```sh
$ pytest --cov --cov-report xml --cov-report term
```

Or if you want to run tests continually whenever a file is changed, like I've been doing while working on this PR, you can install `pytest-watch` and run pytest through it like so:

```sh
$ ptw -- --cov --cov-report xml --cov-report term
```

## Miscellaneous

As much as possible I tried to use the same example inputs/outputs from docstrings as mock data for the tests. Note that in a couple cases this led me to realize the docstring was subtly wrong:

1. camelcase vs. snake_case in `get_available_timefrequency()`
2. frequency_id/source_id order in `GDH()`
3. plural entity types in `get_data_series_list()`


seungatGro (commented):
unrelated to this PR but tests failing for analogous years:

        test_data = get_transform_data.get_data(client, 'metric_id', 'item_id', 'region_id',
>                                               'source_id', 'frequency_id', start_date_bound)

api/client/samples/analogous_years/lib/get_transform_data_test.py:72: 

E       numpy.core._exceptions.UFuncTypeError: ufunc 'subtract' cannot use operands with types dtype('<M8[ns]') and dtype('O')

api/client/gro_client_test.py
```diff
@@ -1,92 +1,373 @@
...
-    }
+def mock_get_allowed_units(access_token, api_host, metric_id, item_id):
+    return [unit['id'] for unit in mock_entities['units'].values()]
 
-    assert client.convert_unit({}, 36) == {}
 
-    with pytest.raises(Exception):
-        assert client.convert_unit({ 'value': 1, 'unit_id': 10 }, 43)
+def mock_get_data_series(access_token, api_host, **selection):
+    return [dict(data_series) for data_series in mock_data_series]
```
seungatGro, line 41: what is the difference between 
[{}, {}]

and

[dict(a) for a in list] WHERE a = {}

some of these mocked functions from lib are just sending response.data, which is a list of dict

api/client/gro_client_test.py
```diff
@@ -1,92 +1,373 @@
...
+
+    def test_get_available(self):
+        self.assertTrue('name' in self.client.get_available('units')[0])
+
+    def test_list_available(self):
+        self.assertTrue('metric_id' in self.client.list_available({})[0])
+
+    def test_lookup(self):
+        self.assertEqual(self.client.lookup('units', 10)['name'], 'kilogram')
+
```
seungatGro, line 194: minro, missing tests for search_and_lookup, lookup_belongs

api/client/gro_client_test.py
```diff
@@ -1,92 +1,373 @@
...
+        client = GroClient(MOCK_HOST, MOCK_TOKEN)
+        client.add_single_data_series(mock_data_series[0])
+        df = client.get_df(show_revisions=True)
+        self.assertEqual(df.iloc[0]['start_date'].date(), date(2017, 1, 1))
+
+    def test_add_points_to_df(self):
+        client = GroClient(MOCK_HOST, MOCK_TOKEN)
+        client.add_points_to_df(None, mock_data_series[0],
+                                client.get_data_points(**mock_data_series[0]))
+        df = client.get_df()
```
seungatGro: minor. use a mock data frame? this fn is also tested in this module

api/client/gro_client_test.py
```diff
@@ -1,92 +1,373 @@
...
+from datetime import date
 
-import pytest
 from api.client.gro_client import GroClient
 
 MOCK_HOST = 'pytest.groclient.url'
 MOCK_TOKEN = 'pytest.groclient.token'
 
-client = GroClient(MOCK_HOST, MOCK_TOKEN)
+mock_entities = {
```
seungatGro: import mock data from module

nemozen (approved):

api/client/gro_client_test.py
seungatGro: in general some of these nested tested client fn should be mocked when being called by another. ie gdh returns client.get_df, are we not doing any stubs or mocking client.get_df for this case. i guess it works for now since these are also tested in this module, but if they require changes, we would need to update two unite tests

seungatGro (changes requested):
minor comments on client tests. missing some coverage, missing some test-specific mocks/stubs

api/client/gro_client_test.py
itsjohnward: ? They're there:

* https://github.com/gro-intelligence/api-client/pull/218/files#diff-dca7837b9521ac58ccaf2591db344cb5R223
* https://github.com/gro-intelligence/api-client/pull/218/files#diff-dca7837b9521ac58ccaf2591db344cb5R229

seungatGro: sorry, mustve misssed them!

api/client/gro_client_test.py
itsjohnward: Using `dict(a)` gives a copy of `a` instead of `a` itself. In `find_data_series()` it calls `get_data_series()` internally and then modifies the output. Before this `dict()` change, that ended up modifying the test data which affected other tests.

api/client/gro_client_test.py
itsjohnward: Yeah, fair. This is blending unit/integration testing. Will split up

api/client/gro_client_test.py
itsjohnward: :+1:

itsjohnward: Done. Left a todo comment about using conftest more to share mock data between different test files (that would let you mock out requests.get() generally and then you don't need specific mocking in test files at all), but for now I just put the mock data in its own module and imported it. Good first step

api/client/gro_client_test.py
itsjohnward: Okay so I've been thinking about this a bit: for `GDH()`, I separated out the parsing logic into its own util now and tested that separately from the `GDH()` function itself. So that's a bit more of a "unit" now that is unit tested.

In terms of mocking `get_df()` in `GDH()` though, I think the tests should know as little as possible about how the methods are implemented - it shouldn't matter if it calls `get_df()` or not, from the test's perspective. I just want a certain output for my given input. What if the method's changed to use something else internally but the output's the same? That's a fine change - maintains backwards compatibility and everything but is just an internal refactor - but my test would then be invalid.

Similarly, with `add_points_to_df()`, I think the expected behavior is that I add the points using this method and then can get the points back by using `get_df()`. If anything about that usage changes, that seems like a compatibility issue. Not sure I want to be inspecting the internal `_data_frame` property, since that's not one of the public interfaces that's guaranteed not to change - `get_df()` is.

alexsherman (approved):
Changes lgtm. I'm also getting a failure in the analagous years test, a different one from Seung ` AttributeError: 'Series' object has no attribute 'to_numpy'` but that's out of the scope of this PR, yes?

itsjohnward:
> Changes lgtm. I'm also getting a failure in the analagous years test, a different one from Seung ` AttributeError: 'Series' object has no attribute 'to_numpy'` but that's out of the scope of this PR, yes?

Yeah there appears to be an issue with Analogous Years in python 3.8 @joysaikat 

Right now our CI's only testing up to 3.7 since shippable does not have an official 3.8 build available yet https://github.com/Shippable/support/issues/5032 and it works fine on 3.7. I suppose we'll need to start using a custom testing image to be able to get the latest versions

api/client/gro_client_test.py
seungatGro: fair, like i said before, we also tested the get_df fn so i think its reasonable to use here. i think theres arguments to both. now, if get_df is updated this might change the validity for both unit tests. im not sure why you mentioned that get_df is guaranteed to not change.

seungatGro (approved):

PR #220, opened 2020-05-01
add support for retrieving descendants' geojson in a batch

nemozen:


api/client/lib.py
```diff
@@ -598,14 +598,19 @@ def get_geo_centre(access_token, api_host, region_id):
 
 
 @memoize(maxsize=None)
```
itsjohnward, line 600: Need to move the memoize decorator down to the integer `get_geojson()` one

Memoization doesn't work with functions that take a list as their input.

api/client/lib.py
```diff
@@ -598,14 +598,19 @@ def get_geo_centre(access_token, api_host, region_id):
...
+        region_id)])
+    if descendant_region_level_id:
+        url += "&reqRegionLevelId={}&stringify=false".format(descendant_region_level_id)
     headers = {'authorization': 'Bearer ' + access_token}
     resp = get_data(url, headers)
-    for region in resp.json()['data']:
+    return [dict_reformat_keys(r, str_camel_to_snake) for r in resp.json()['data']]
+
+
+def get_geojson(access_token, api_host, region_id):
```
itsjohnward, line 612: Memoize decorator on this one as well?

api/client/lib.py
```diff
@@ -598,14 +598,19 @@ def get_geo_centre(access_token, api_host, region_id):
 
 
 @memoize(maxsize=None)
-def get_geojson(access_token, api_host, region_id):
-    url = '/'.join(['https:', '', api_host, 'v2/geocentres?includeGeojson=True&regionIds=' +
-                    str(region_id)])
+def get_geojsons(access_token, api_host, region_id, descendant_region_level_id=None):
```
itsjohnward: I was confused at first by this name because I thought it implied region_id could be a list.

I see now it's still an integer but you also specify region_level to get multiple results.

The endpoint does support lists of regionIds as inputs though, so if you wanted to do `get_geojsons([region['id'] for region in get_descendant_regions()])` you could do that as well, if the function supported it.

But right now that breaks on the `memoize` decorator, since you can't hash lists as an input

itsjohnward (changes requested):

api/client/gro_client.py
```diff
@@ -302,8 +302,24 @@ def get_geo_centre(self, region_id):
         """
         return lib.get_geo_centre(self.access_token, self.api_host, region_id)
 
+
+    def get_geojsons(self, region_id, descendant_region_level_id=None):
```
itsjohnward: Could we make the `descendant_region_level` kwarg name consistent with `get_descendant_regions`?

`descendant_level` is what it uses

itsjohnward (changes requested):

api/client/lib.py
nemozen: not memoizing here was intentional, since this fn calls the other ,there's no benefit memoizing the middle man, and the cost is 2x the ram

api/client/gro_client.py

api/client/lib.py
nemozen: prefer to avoid ambiguous types in fn args, "int or list"  is an anti-pattern (ok for short term compatibility/deprecation/hacks :).  In general better to give the user the choice explicitly, singular or plural and only have the choice when necessary.  For this there is no current use case for mutliple region trees, i think it's efficient enough to do one tree at a time, since the root could be the whole world or a whole continent etc..  o

api/client/gro_client.py
itsjohnward: This would be a breaking change, changing the name in `get_descendant_regions`

I meant change `get_geojsons` since it's an addition and nothing depends on it yet

itsjohnward (changes requested):

api/client/lib.py
itsjohnward: Ah I see. 👍

api/client/lib.py
itsjohnward: Since it's a new function, I was thinking it could just require a list as an input, in contrast to the function we already have that takes an int.

If you only want one region, you could always just do `[region_id]`, but if we want to allow lists in the future now it'd have to be a 3rd separate "root" function that these two call. Getting deeper

api/client/gro_client.py
nemozen: right it's a kwarg not an arg! done

api/client/lib.py
nemozen: right but don't think there's a good use case, so odn't want to encourage unnecessarily large calls.   The real use case is process a region tree, there's n advantage to processing multiple trees in one batch, sinc ethe tree can be th ewhole world or a continent.

itsjohnward (approved):

api/client/gro_client.py
```diff
@@ -302,8 +302,24 @@ def get_geo_centre(self, region_id):
         """
         return lib.get_geo_centre(self.access_token, self.api_host, region_id)
 
+
+    def get_geojsons(self, region_id, descendant_level=None):
+        """Given a region ID, return shape information in geojson, for the
+        region and all its descendants at the given level (if specified).
```
seungatGro, line 308: add valid list of descendant_levels. admin levels 2~5. getting api response error:
`{"statusCode":400,"error":"Bad Request","message":"child \"reqRegionLevelId\" fails because [\"reqRegionLevelId\" must be less than or equal to 5]","validation":{"source":"query","keys":["reqRegionLevelId"]}}`

seungatGro (changes requested):
minor change on input params for descendant_level.

otherwise lgtm

api/client/gro_client.py
nemozen: done (although we should generalize it on the server side to include 8)

PR #221, opened 2020-05-02
return list instead of generator in get_geojsons

nemozen:


itsjohnward (approved):

api/client/lib.py
```diff
@@ -605,8 +605,7 @@ def get_geojsons(access_token, api_host, region_id, descendant_level=None):
         url += "&reqRegionLevelId={}&stringify=false".format(descendant_level)
     headers = {'authorization': 'Bearer ' + access_token}
     resp = get_data(url, headers)
-    for r in resp.json()['data']:
-        yield dict_reformat_keys(r, str_camel_to_snake)
+    return [dict_reformat_keys(r, str_camel_to_snake) for r in resp.json()['data']]
```
itsjohnward, line 608: Just checked the docstring to see if it was out of date, but it said returns a list in the first place: https://github.com/gro-intelligence/api-client/pull/220/files#diff-603e79178b8e522d0e366a9fca477b24R315

PR #223, opened 2020-05-07
CLEWS-24276: Allow mixed unit series in get_data_points

itsjohnward:
Datapoint output looks something like in the docstring:

```py
[{
    'series': { ... 'unitId': 4, ... },
    'data': [
        [startDate, endDate, value, reportingDate, unitId, metaData],
     ]
}]
```

This PR switches over from using the one in the `series` object to the one in the `data` array. That way, if the unit is inconsistent within the series, we get the accurate one for that point.

Just to be extra safe, I also added a fallback to the `series` unitId, but that shouldn't be needed. Can be phased out entirely in time. Left in the server output for backwards compatibility.

### DIFF
--- a/api/client/lib.py
+++ b/api/client/lib.py
@@ -437,7 +437,7 @@ def list_of_series_to_single_series(series_list, add_belongs_to=False, include_h
     ...     'data': [
     ...         ['2001-01-01', '2001-12-31', 123],
     ...         ['2002-01-01', '2002-12-31', 123, '2012-01-01'],
-    ...         ['2003-01-01', '2003-12-31', 123, None, {}]
+    ...         ['2003-01-01', '2003-12-31', 123, None, 15, {}]
     ...     ]
     ... }], True) == [
     ...   { 'start_date': '2001-01-01',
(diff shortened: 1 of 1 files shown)


api/client/lib.py
```diff
@@ -505,12 +505,10 @@ def list_of_series_to_single_series(series_list, add_belongs_to=False, include_h
...
                 'end_date': point[1],
                 'value': point[2],
-                # list_of_series has unit_id in the series attributes currently. Does
-                # not allow for mixed units in the same series
-                'unit_id': series['series'].get('unitId', None),
+                'unit_id': point[4] if len(point) > 4 else series['series'].get('unitId', None),
                 # input_unit_id and input_unit_scale are deprecated but provided for backwards
                 # compatibility. unit_id should be used instead.
-                'input_unit_id': series['series'].get('unitId', None),
+                'input_unit_id': point[4] if len(point) > 4 else series['series'].get('unitId', None),
```
seungatGro, line 511: should this still be used? seems incorrect, we can remove since unit is present in each data row?

seungatGro (changes requested):
input_unit_id is not equal to unit_id

api/client/lib.py
itsjohnward: No, see the comment right above:

> input_unit_id and input_unit_scale are deprecated but provided for backwards compatibility. unit_id should be used instead.

In the original versions of the client, we only had `input_unit_id` and `input_unit_scale` since there was no server-side conversion happening at all. When we started doing some conversions, `unit_id` was added to be accurate, but anyone using `input_unit_id` and `input_unit_scale` would've started getting incorrect values since it no longer reflected what unit the value was in.

So those two values are set to the unit_id and 1 so that legacy code will continue to function as expected. It's not documented, and it's discouraged. It's one of the few accumulated cruft things slated for removal in an eventual Major version update, but no timeline on that just yet.

seungatGro (approved):

nemozen:
lgtm

PR #224, opened 2020-05-18
CLEWS-27625 Improve BatchClient test coverage

itsjohnward:
Pre-req: https://github.com/gro-intelligence/api-client/pull/218

### DIFF
--- a/api/client/batch_client.py
+++ b/api/client/batch_client.py
@@ -145,8 +145,9 @@ class BatchClient(GroClient):
         if show_revisions:
             for data_series in self._data_series_queue:
                 data_series['show_revisions'] = True
-        self.batch_async_queue(self.get_data_points, self._data_series_queue, None,
-                               self.add_points_to_df)
+        self.batch_async_get_data_points(self._data_series_queue,
+                                         output_list=self._data_frame,
+                                         map_result=self.add_points_to_df)
         return self._data_frame
 
     # TODO: deprecate  the following  two methods, standardize  on one
--- a/api/client/batch_client.py
+++ b/api/client/batch_client.py
@@ -132,15 +132,6 @@ class BatchClient(GroClient):
         # Retries failed. Raise exception
         raise BatchError(response, retry_count, url, params)
 
-    @gen.coroutine
-    def get_data_points(self, **selection):
-        """Get all the data points for a given selection, which is some or all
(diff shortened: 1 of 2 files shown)


api/client/batch_client.py
```diff
@@ -132,21 +132,13 @@ def log_request(start_time, retry_count, msg, status_code):
...
             for data_series in self._data_series_queue:
                 data_series['show_revisions'] = True
-        self.batch_async_queue(self.get_data_points, self._data_series_queue, None,
-                               self.add_points_to_df)
+        self.batch_async_get_data_points(self._data_series_queue,
+                                         output_list=self._data_frame,
+                                         map_result=self.add_points_to_df)
         return self._data_frame
 
     # TODO: deprecate  the following  two methods, standardize  on one
```
alexsherman, line 144: If we are removing `get_data_points` and its usage in `get_df`, is this TODO comment about deprecating `batch_async_get_data_points` relevant anymore?

api/client/batch_client_test.py
```diff
@@ -1,10 +1,184 @@
...
+        list_of_ranked_series_lists = self.client.batch_async_rank_series_by_source(
+            [mock_data_series, mock_data_series]
+        )
+        # There were two inputs, so there should be two outputs:
+        self.assertEqual(len(list_of_ranked_series_lists), 2)
+        for series_list in list_of_ranked_series_lists:
+            # Not necessarily true, but true given the mock_rank_series_by_source() function:
+            self.assertEqual(len(series_list), len(mock_data_series))
+            for series in series_list:
+                self.assertTrue('metric_id' in series)
```
alexsherman, line 183: Why assert that metric is in series and not other entity types?

alexsherman (approved):
LGTM, just a few minor questions. Only tested on python 3

api/client/batch_client_test.py
itsjohnward: Could assert all of them, but I just picked metric_id randomly as a test set. Source is the important one, plus one "other"

api/client/batch_client.py
itsjohnward: Yeah I think that'll be taken care of when BatchClient and GroClient are merged together. I just want one `get_data_points()` function that does both sync and async versions, depending on what the input is and whether the environment supports async. User shouldn't have to care about the difference

PR #225, opened 2020-05-26
Update batch_async_get_data_points test to use point unit

itsjohnward:
https://github.com/gro-intelligence/api-client/pull/223 was merged but never pulled into https://github.com/gro-intelligence/api-client/pull/224 so when the latter was merged to dev, this test failed: https://app.shippable.com/github/gro-intelligence/api-client/runs/3223/1/console

Now because of https://github.com/gro-intelligence/api-client/pull/223, the unit id is being read from the point instead of from the series, by default. So in our `mock_list_of_series_points` variable we have two different ones:

```py
['2017-01-01T00:00:00.000Z', '2017-12-31T00:00:00.000Z', 40891, None, 14, {}],
['2018-01-01T00:00:00.000Z', '2018-12-31T00:00:00.000Z', 56789, '2019-03-14T00:00:00.000Z', 10, {}],
```

14 and 10. Old way was it was reading the 14 from

```py
'series': {
  'unitId': 14,
  ...
}
```

and ignoring the 10 on the point.

### DIFF
--- a/api/client/batch_client_test.py
+++ b/api/client/batch_client_test.py
@@ -16,55 +16,14 @@ from tornado.ioloop import IOLoop
 
 from api.client.batch_client import BatchClient, BatchError
 from api.client.utils import str_camel_to_snake
+from api.client.mock_data import (
+    mock_list_of_series_points,
+    mock_data_series,
+    mock_error_selection,
+)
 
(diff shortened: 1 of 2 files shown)


alexsherman (approved):
LGTM!

PR #226, opened 2020-05-28
Merge BatchClient functions into GroClient

itsjohnward:
1. Delete `api/client/batch_client.py` and leave an import alias to GroClient in its place
2. Add all the batch_async and tornado methods into GroClient

All GroClient and BatchClient tests still pass, the only modification being I needed to separate `get_df()` and `async_get_df()`. If you were using the old `BatchClient.get_df()` then the new `GroClient.get_df()` should work but may slow down.


itsjohnward:
I separated out the last few commits into their own PR https://github.com/gro-intelligence/api-client/pull/231 since they were just aiming to increase test coverage.

api/client/gro_client.py
```diff
@@ -7,19 +7,64 @@
...
+            # For example, a 400 Bad Request might say "metricId is required"
+            if "message" in json_content:
+                self.message += ": {}".format(json_content["message"])
+        except Exception:
+            # If the error message can't be parsed, fall back to a generic "giving up" message.
+            self.message = "Giving up on {} after {} {}: {}".format(
+                self.url,
+                self.retry_count,
+                "retry" if self.retry_count == 1 else "retries",
+                response,
```
cn1036, line 64: minor: extra comma

cn1036:
I was testing the sample script _batch_queries.py_ but the error handling is a bit confusing. we are not raising error if some of the requests fail? output is a list of BatchErrors. As you can see, it continues to process the error as valid data
`output = api_client.batch_async_get_data_points(selections)`
image

api/client/gro_client.py
itsjohnward: That's actually an intended code style thing: https://black.readthedocs.io/en/stable/the_black_code_style.html#trailing-commas

See https://stackoverflow.com/a/11597911 for some justification of why that's a good thing

itsjohnward:
> I was testing the sample script _batch_queries.py_ but the error handling is a bit confusing. we are not raising error if some of the requests fail? output is a list of BatchErrors. As you can see, it continues to process the error as valid data
> `output = api_client.batch_async_get_data_points(selections)`
> image

Yeah `batch_async_get_data_points` passes errors on to you as the user to determine how to handle them. Can see the description of https://github.com/gro-intelligence/api-client/pull/177 for a few examples of failing fast vs. ignoring errors. Also can see [here](https://github.com/gro-intelligence/api-client/blob/e7d0e637ef7ead343a98d69fafbab84832164aaa/api/client/batch_client_test.py#L124) where that behavior's explicitly tested. If you're processing 1000s of requests and just 1 of them failed, it depends what your use case is, whether you want to keep retrying that 1 or if 99.9% coverage is acceptable and you want to just ignore it. With this error propagation the idea is you can determine which is best for you.

That sample script doesn't do any special error handling and assumes the requests will succeed. Could be more explicit about what it's supposed to do on error, but since it's a toy example I think there's some value in not making it unnecessarily complicated. It fails in an okay way as it is, showing you the 402 error that's causing the issue.

And that's out of the scope of this PR either way, since that error handling and the sample script both have not been modified in the migration of `batch_async_get_data_points` from BatchClient over to GroClient. Should behave identically before and after.

cn1036:
> > I was testing the sample script _batch_queries.py_ but the error handling is a bit confusing. we are not raising error if some of the requests fail? output is a list of BatchErrors. As you can see, it continues to process the error as valid data
> > `output = api_client.batch_async_get_data_points(selections)`
> > image
> 
> Yeah `batch_async_get_data_points` passes errors on to you as the user to determine how to handle them. Can see the description of #177 for a few examples of failing fast vs. ignoring errors. Also can see [here](https://github.com/gro-intelligence/api-client/blob/e7d0e637ef7ead343a98d69fafbab84832164aaa/api/client/batch_client_test.py#L124) where that behavior's explicitly tested. If you're processing 1000s of requests and just 1 of them failed, it depends what your use case is, whether you want to keep retrying that 1 or if 99.9% coverage is acceptable and you want to just ignore it. With this error propagation the idea is you can determine which is best for you.
> 
> That sample script doesn't do any special error handling and assumes the requests will succeed. Could be more explicit about what it's supposed to do on error, but since it's a toy example I think there's some value in not making it unnecessarily complicated. It fails in an okay way as it is, showing you the 402 error that's causing the issue.
> 
> And that's out of the scope of this PR either way, since that error handling and the sample script both have not been modified in the migration of `batch_async_get_data_points` from BatchClient over to GroClient. Should behave identically before and after.

The code of merging BatchClient looks good to me! approving

cn1036 (approved):

PR #227, opened 2020-06-01
New version of similar regions

ash-gro:
Full rewrite of 'similar regions' based on different metric (Frechet, a.k.a. 2-Wasserstein) treating time series data as a distribution. Tested on AWS unix node under python 3.6.9 only.

 All code is in samples/similar_regions_Frechet samples/sr.py with metric.py giving API d/l details. See samples/similar_regions_Frechet samples/test.py for usage examples as well as some d/l failure messages ("Connection closed" and "502 Bad Gateway") which would ideally need to be addressed.

Warning: full data download takes many hours. Contact MA if you want to get some already downloaded data which should be sufficient to test region comparison (as opposed to data download) functionality. It is ~57M, so can not be attached here.

Separately, a few lines were added to batch_client.py to fix what I believe is a bug in handling of certain types of downloading errors


api/client/batch_client.py
```diff
@@ -122,8 +122,10 @@ def log_request(start_time, retry_count, msg, status_code):
                     if retry_count > 0:
                         time.sleep(2 ** retry_count)
                     continue
-                elif status_code in [400, 401, 402, 404]:
+                elif status_code in [400, 401, 402, 404, 502]:
```
itsjohnward, line 125: 502 should be retryable. Could put in the other list of status codes above

itsjohnward (dismissed):

api/client/batch_client.py
ash-gro: Shoult be done now

api/client/samples/similar_regions_Frechet/sr.py
```diff
@@ -0,0 +1,560 @@
+import os, glob
```
eliaschneider, line 1: maybe call this file similar_region.py, so it's more clear that it's main file

api/client/samples/similar_regions_Frechet/sr.py
```diff
@@ -0,0 +1,560 @@
...
+                 search_region=0,
+                 update_mode="no",
+                 data_dir=None,
+                 region_info_reload=False,
+                 t_int_per_year=52,
+                 metric_instance=None,
+                 drop_mode='any_missing'):
+        """
+        :param metric_properties: A dict containing properties which can be used in the region similarity metric.
+        This is by default defined in mtric.py, but can be adjusted.
```
eliaschneider, line 44: typo mtric instead of metric

api/client/samples/similar_regions_Frechet/metric.py
```diff
@@ -0,0 +1,179 @@
+### This file describes the metric to be used when doing region similarity computation.
```
eliaschneider, line 1: This file seems an example sample, we may want to indicate that in the filename

api/client/samples/similar_regions_Frechet/metric.py
```diff
@@ -0,0 +1,179 @@
...
+# Each data point belonging to an item will be MULTIPLIED by sqrt of the weight
+
+# Full calculation before applying the distance measure is therefore sqrt(weight)*data/norm.
+# Note that taking sqrt of user-provided weights here assumes user wants something like
+# dist^2 = \sum w_user_i*(x_i - y_i)^2 (note non-squared w_user_i here)
+# which is usual specification in weighted L2 norm (least squares)
+
+
+# Just an example. This is intended to be define in the user code
+metric_instance = {
```
eliaschneider, line 29: should we include weight in the variable name?

api/client/samples/similar_regions_Frechet/sr.py
```diff
@@ -0,0 +1,560 @@
...
+        :param search_region: region_id to search for similar regions. Default is 0 (entire world).
+        :param update: If 'reload', full reload from Gro API after which local caches are completely overwritten
```
eliaschneider, line 46: here it should be update_mode

api/client/samples/similar_regions_Frechet/sr.py
```diff
@@ -0,0 +1,560 @@
...
+            try:
+                os.mkdir(CACHE_PATH)
+                self.data_dir = CACHE_PATH
+            except:
+                # last resort
+                self.data_dir = tempfile.gettempdir()
+        
+        self._logger.info("*********Creating similar region object with *******\n\tregion {}\n\tupdate_mode {}\n\tdata_dir {}\n\tregion_info_reload {}\n\tt_int_per_year {}\n".format(search_region,update_mode,self.data_dir,region_info_reload,t_int_per_year))
+
+        self.client = BatchClient(API_HOST, ACCESS_TOKEN)
```
eliaschneider, line 92: you can use directly GroClient, BatchClient is exactly the same with last Gro api release

api/client/samples/similar_regions_Frechet/sr.py
```diff
@@ -0,0 +1,560 @@
...
+                if valid_regions:
+                    np.savez_compressed(prop_path,
+                                data = np.concatenate([prop_data_cached,valid_data],axis=0),
+                                regions = prop_regions+valid_regions)
+                    self._merge_into_prop_data(valid_data, valid_regions)
+                    
+        ################### Drop incomplete regions #################################
+        self.data = pd.DataFrame(self.data,
+                            index=pd.MultiIndex.from_product([self.needed_regions,range(1,self.t_int_per_year+1)],
+                                                             names=['region','month']),
```
eliaschneider, line 171: instead of 'month' maybe 'year_period'?

api/client/samples/similar_regions_Frechet/sr.py
ash-gro: This IS the main file (the analog of similar_regions.py AND similar_region_state.py put together in the original version). similar_regions_Frechet is fully self-contained.

api/client/samples/similar_regions_Frechet/sr.py
ash-gro: fixed

api/client/samples/similar_regions_Frechet/metric.py
ash-gro: I guess one could say this is an example. However, some sort of metric description is required to run the thing and in most cases it will created from a subset of what is presented in this file (rather than typed from scratch), I am reluctant to change the file name.

ash-gro: Above I am talking mostly about metric_description which is the second (longer) dictionary in this file. I guess one could split this into two but I don't like small file proliferation. Can still change if you have a strong opinion on this...

api/client/samples/similar_regions_Frechet/metric.py
ash-gro: Probably not, at least not yet. While it is true that right now this only contains property weights, I actually moved things back and forth between metric_instance and metric_description and still not sure what the best split is. For example, "norm" might be better suited for metric_instance...

api/client/samples/similar_regions_Frechet/sr.py
ash-gro: fixed

api/client/samples/similar_regions_Frechet/sr.py
ash-gro: If you say so (don't think it was the case when I started this). But I guess it then does not hurt to keep BatchClient (i.e. as it is now) either unless we plan to deprecate it which I assume we don't. Keeping as is should avoid the need for new testing.

api/client/samples/similar_regions_Frechet/sr.py
ash-gro: self.data exists as dataframe for a short time (goes back to numpy array quickly), but sounds good, changed.

api/client/samples/similar_regions_Frechet/sr.py
```diff
@@ -0,0 +1,560 @@
...
+PICKLE_PROTOCOL = 4
+REGIONS_PER_QUERY = 100
+MAX_RETRY_FAILED_REGIONS = 3
+OK_TO_PROCEED_REGION_FRACTION = 0.1 # we want at least 10% of all desired region present in the final data matrix
+
+class SimilarRegion(object):
+
+    def __init__(self, metric_properties,
+                 search_region=0,
+                 update_mode="no",
```
nemozen, line 36: i think we have too many options here.   Lets' just make it load a pickle file with the whole object and recover a prior state.  so basically replace update_mode, and region_info_reload and data_dir with just a file path.   If the file is not given  it just starts empty. if the file is given but doesn't exist it should  fail.

api/client/samples/similar_regions_Frechet/sr.py
```diff
@@ -0,0 +1,560 @@
...
+                to_include = [r for r in [j for r in to_include for j in self.region_info[r]['contains']] 
+                              if self.region_info.get(r,{'level':-1})['level']==level+1]
+                ri.update({r:self.region_info[r] for r in to_include})
+                level += 1
+            self.region_info = ri
+            self._logger.info("Trimmed region info to {} items".format(len(self.region_info)))
+        self.needed_regions = sorted(self.region_info.keys())
+        self.needed_properties = sorted(self.metric_instance.keys())
+        
+        ##################### Data download ################################################
```
nemozen, line 134: let's make each of these clearly separated sections a function call... 

-  self. _initialize_regions()
-  self._data_download()
-  self._normalize()
etc.

api/client/samples/similar_regions_Frechet/sr.py
nemozen: done in 73d773ddab14619279ab7152c90922d827439949

i've simplified it so it always uses the cache if present and updates if anything needed is missing, both region info and data. the user also can just delete the cache or point to a new  dir to force reload, which is also convenient when working on different scenarios

api/client/samples/similar_regions_Frechet/sr.py
```diff
@@ -0,0 +1,560 @@
...
+                for datapoint in response:
+                    v = datapoint["value"]
+                    
+                    # No action on N/A
+                    if v is not None:
+                        # Dates are recieved as starting from YYYY-MM-DD
+                        # (for example 'start_date': '2017-01-01T00:00:00.000Z')
+                        # ugly date conversion, but this is inner loop, strptime is too slow here
+                        # TODO: python 3.7 has fromisoformat which is supposedly much faster than strptime
+                        y = int(datapoint["start_date"][:4])
```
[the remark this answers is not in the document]
ash-gro: I tried several pandas calls here (although not 100% sure where this included exactly the one you propose) and they were all many times slower than the simple string slice followed by int conversion. So let's keep as is for now. Might revisit when we have python 3.7 as default

api/client/samples/similar_regions_Frechet/sr.py
ash-gro: For "single-use" code I personally do prefer a single stream of code with comments rather than a sequence of calls to single-use functions but yes, this is somewhat contrary to current software engineering "wisdom", so as you wish ...

api/client/samples/similar_regions_Frechet/sr.py
ash-gro: Maybe I am missing something, but this sounds like pickling a (previously created) SR object. But we don't need this functionality within SR class - just do pickle.dump/pickle.load as with any other class. And some sort of "partial pickling" is probably an unnecessary complication.

 I agree that  there are quite a few parameters here, but this is a complex object and at some point it needs to be generated from scratch, so even if we pickle most of the object somehow, we still need to specify full configuration from time to time. However, for a "casual" user defaults should be fine. In fact, maybe we can remove region_info_reload and drop_mode altogether (always use their default values).

 I'd prefer to keep update_mode - "starting empty" means full reload and, given the time it takes, this should be requested explicitly (via update_mode="reload").  data_dir is mostly to have the ability to work with multiple data caches. This is convenient sometimes, but might not be needed by a typical user, so can be dropped as well (i.e. we will always use ./sr_cache directory). We can go even a step further and kill t_int_per_year option (always use 52) since users probably not going to play with this (and it required data re-download). 
 
With all these together we will have minimal form as something like
    def __init__(self, metric_properties,
                 search_region=0,
                 update_mode="no",
                 metric_instance=None)
In fact, with only this few parameters, data_dir could probably be kept as well as it does come convenient sometimes.

Overall proposal is therefore to remove only region_info_reload, t_int_per_year, and drop_mode. I think this should be sufficient. Note that these options can still be accessed via globals in the code. I guess we expect advanced users to be able to modify code if needed (since we are releasing the code in the first place)

api/client/samples/similar_regions_Frechet/sr.py
nemozen: A good rule of thumb is that you should be able to read the whole function without too much scrolling, so you can understand what it does and not miss any side effects etc. It's also good to be able to unit test each logical block.  Single use becomes n-uses when you want to test :)

api/client/samples/similar_regions_Frechet/sr.py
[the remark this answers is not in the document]
nemozen: ok

api/client/samples/similar_regions_Frechet/sr.py
nemozen: My earlier comment was a bit to succinct... i did keep data_dir and the separate region and date cache files as before.  Here's what we have now
- always loads saved region info  if it finds it in the specified data_dir
- always updates as necessary if region is missing
- load data if any is missing by default  (but still supports the other "drop_modes" since it's already done)

Also as a bonus, I've made it so that

-  supports comparison across different region trees.. E.g. you can pick a district in US and find similar provinces in Europe.
-  when the comparison regions changed, it is automatically updated. so basically the user only needs to call similar_to() no matter what the regions are and never worry about re-initializing.

nemozen:
> code org: we don't want to create another module, i think one is enough at least by default. but if we want to save the old metric, they can coexist as two different options to load. so let;s move similar_regions_Frechet.sr.SimilarRegion to similar_regions.similar_region_frechet.SimilarRegion and we'll have two top-level classes
> 
> testing in progress

all changes done  except moving the module.

nemozen:
ok made a few changes to improve performance 

only get necessary regions (this is 10x faster when the user only wants province level)
https://github.com/gro-intelligence/api-client/pull/227/commits/b3ea3b6f5490fb6e3c56957c1f1c92bacd00d86a

 only retry on api errors and exponentially smaller batch sizes  when retrying 
https://github.com/gro-intelligence/api-client/pull/227/commits/fe5cad980f3176892b37fe95749a9bc7321a0f5e
https://github.com/gro-intelligence/api-client/pull/227/commits/e115a19daf5be7770f3eace3adc2e522772adaca

also cleanup caching so it always uses existing cache and the user can still control it explicitly more simply by having different cache dir
https://github.com/gro-intelligence/api-client/pull/227/commits/5853ff3ba8218d26ba7f488a0fbeb5f19e78c0ad

using new GroClient location
https://github.com/gro-intelligence/api-client/pull/227/commits/1e0a5bcf34bd7e6dc1d3341442082a49dd097e42

retested and updated examples, looks good 

next step will merge into the  similar_regions module

api/client/samples/similar_regions_Frechet/metric.py
nemozen: renamed to metric_weights

api/client/samples/similar_regions_Frechet/sr.py
nemozen: done

nemozen:
merged new version into existing  similar_regions module
fccb32d0f24170362ad348ad80961dbe6cc9733e

nemozen (approved):

api/client/batch_client.py
nemozen: moved this change to https://github.com/gro-intelligence/api-client/pull/285

nemozen:
> I still strongly dislike not having explicit user-controlled "update_mode" parameter. It seems that the modified version will behave similar to update_mode='update' in the original. The problem is that often one knows that there has been no updates to the database (for example, we just ran similar regions a second ago!) and local cache is complete even if there are some data points missing. The new version will always go to database if there are missing data which can be quite wasteful. Complete cache recreation option (update_mode='reload' in the original) can of course be simulated by manually deleting cache files/directory but that seems less convenient

re: "reload", yeah.. note that you don't have to delete it, you can just point to a different location. I prefer not to have two ways to do things, (boolean and name of dir)

re: "no update", I considered that but in practice, checking for missing data is much better because
-  the data can be filled in,  not just due to data quality updates from gro or the source which I agree would not change in minutes, but also more commonly if you have client side or server side network issues  where some data  points were missed despite the retries, then it does help to refill it.
- In terms of time, the cost is negligible.    if you are missing a hundred regions it's only a  4 http requests in parallel so 1 round trip time. The absolute worst case I've seen is still  less 10 seconds , and the overall performance is dominated by the time to  to read the cache from disk (a couple of minutes)

jli (approved):
this change lgtm, though note ash's comment about ignoring other status codes (also if the exception didn't have the `code` attribute, the loop will continue)

nemozen:
> Removing region info caching is ok - I was starting this when get_descendant_regions was very slow.
> 
> Rebuild is triggered only if new search region is different from self.search_region (line 585), but, as is, it will NOT be triggered on requested level mismatch. So something incorrect will likely happen if one first searches over provinces and then over districts of the same region as in
> similar_to(region_id, requested_level=4) # this will build search trees for level 4 only
> similar_to(region_id, requested_level=5) # searching over the same region, so no object rebuilding, and no search tree for level 5
> (since data structures are now only build for requested levels). Needs to be fixed with some tracking/checking of current level(s) on which data structures exist
> 

Yes! Meant to add level to the tracking but forgot. Done.

> Finally, I think returning all parents (and respectively grandparents) is very inconvenient for the user. Many of these parents are very obscure region collections (something like "Ag region 42"), so this will generate potentially a lot of info nobody cares about within which it will be hard to find what is most needed. For example, for a US county parent/grandparent should just be US_state, USA. So at the very least, only return parents/grandparents on correct levels. For 3/4/5 level regions these are expected to form a tree, so it should be sufficient to return just one.

Done.

FYI, I do want to add region level 8 as well so it works for non-admin regions that are still interesting as physical regions and therefore have geospatial data e.g. water basins, agricultural zones etc.   Also in many cases, we expect the user to add more metadata and formatting outside of the sr core logic,  e.g. to draw a map you would get the geojson. Will add an  examples to the notebook separately. So we don't need to add too much to the output here in the core logic, but picking the most important parent and grandparent (the admin parent) is basic enough that it's worth it.

nemozen:
added a few improvements to README and example notebook

nemozen:
final review done offline with @ash-gro

PR #229, opened 2020-06-09
Add include_geojson and zoom_level to the params of get_geo_centre()

cn1036:
Status: tested locally, ready for review.

API users will be able to specify whether shapefile needs to be included in the response.
API endpoint changes: https://grointelligence.atlassian.net/browse/CLEWS-27183


nemozen (changes requested):
we need zoom level support on get_geojson() not get_geo_centre()

cn1036:
@nemozen 
sorry, working on stage RC right now. will review the commit and update unit tests soon

api/client/gro_client.py
```diff
@@ -301,38 +301,48 @@ def get_geo_centre(self, region_id):
...
             Example::
 
-                [{'centre': [ 45.7228, -112.996 ], 'regionId': 1215, 'regionName': 'United States'}]
-
+                [{'centre': [ 39.8333, -98.5855 ], 'regionId': 1215, 'regionName': 'United States'}]
         """
         return lib.get_geo_centre(self.access_token, self.api_host, region_id)
 
-    def get_geojsons(self, region_id, descendant_level=None):
+    def get_geojsons(self, region_id, descendant_level=None, zoom_level=7):
```
cn1036, line 656: just to confirm, we used to return simplified shapefile for all requests. is it intended to change the default to detailed boundaries table? @nemozen

nemozen:
Yes, we want to keep the current server-side default, so we don't change
client side performance of the web app. for api clients, we want to default
to more detailed boundaries (as with shp files), and give  the client the
option to reduce the zoom level if higher level of accuracy is not needed.

api/client/gro_client.py
nemozen: Yes, we want to keep the current server-side default, so we don't change
client side performance of the web app. for api clients, we want to default
to more detailed boundaries (as with shp files), and give  the client the
option to reduce the zoom level if higher level of accuracy is not needed.

nemozen (approved):

itsjohnward:
^ I see the shippable run just failed on "pluggy.manager.PluginValidationError: Plugin 'pytest_cov' could not be loaded"

That error should be fixed by https://github.com/gro-intelligence/api-client/pull/232. Merging from dev to re-run tests.

api/client/lib.py
```diff
@@ -538,26 +538,26 @@ def lookup_belongs(access_token, api_host, entity_type, entity_id):
...
-    resp = get_data(url, headers)
+    resp = get_data(url, headers, {'regionIds': region_id})
     return resp.json()['data']
 
 
 @memoize(maxsize=None)
-def get_geojsons(access_token, api_host, region_id, descendant_level=None):
-    url = '/'.join(['https:', '', api_host, 'v2/geocentres?includeGeojson=True&regionIds={}'.format(
-        region_id)])
+def get_geojsons(access_token, api_host, region_id, descendant_level, zoom_level):
```
itsjohnward, line 548: These lib functions, `get_geo_centre`, `get_geojsons`, and `get_geojson` are not covered by any tests, and since this PR is adding code to them, it has decreased test coverage a bit. Add these to `lib_test.py`?

itsjohnward (changes requested):

api/client/lib.py
cn1036: yeah you're right. working on it now

PR #231, opened 2020-06-12
Increase find_data_series test coverage and fix start_date/end_date bug

itsjohnward:
Refactoring:

1. I don't know of any instance in which `if "end_date" in tmp.columns:` etc. would not be true, unless the dataframe was empty, which is already handled above. Removed them because it seems like an unnecessary conditional.
2. Giving `result_filter` a default value of `None` eliminates a little bit of complexity, not having to pop off of `kwargs`

Bug fixes:

1. `source_id` and `source_name` were being popped off of selections before going into ranking, but only `frequency_id` was. So `frequency_name` could still get out of sync. Removed `frequency_name` as well. Related: https://github.com/gro-intelligence/api-client/pull/185#discussion_r396763536
2. `start_date` and `end_date` inputs to `find_data_series` were broken because it attempts to get the entity type of the key `start_date_id`/`end_date_id` [here](https://github.com/gro-intelligence/api-client/pull/231/files#diff-603e79178b8e522d0e366a9fca477b24R834) to pass into search. That results in a `KeyError`

### DIFF
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -542,13 +542,11 @@ class GroClient(object):
         # get_data_points response doesn't include the
         # source_id. We add it as a column, in case we have
         # several selections series which differ only by source id.
-        tmp["source_id"] = data_series["source_id"]
-        if "end_date" in tmp.columns:
-            tmp.end_date = pandas.to_datetime(tmp.end_date)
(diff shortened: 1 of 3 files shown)


eliaschneider (approved):
LGTM

PR #232, opened 2020-06-12
CLEWS-28861: lock pytest-cov version at 2.9.0

itsjohnward:
pytest-cov 2.10.0 increased the minimum pytest version from 3.6 to 4.6: https://github.com/pytest-dev/pytest-cov/compare/v2.9.0...v2.10.0#diff-2eeaed663bd0d25b7e608891384b7298R125

We're currently resolving to 4.2.1, so we no longer meet the minimum version requirement. Rolling back to 2.9.0 as a quick fix to see if that allows tests to go through. Follow-up should be to figure out what's causing us to stop at pytest 4.2.1.

Pytest 5.0.0 removes Python 2 support, but based on python_requires we should still be able to go up to pytest 4.6.11.


eliaschneider (approved):
LGTM

PR #236, opened 2020-07-04
Add entity names to get_df index_by_series

stephenbernardi:
To finish up CLEWS-15676, this PR converts entity ids to names when index_by_series is chosen.

In addition, another option (index_by_end_date) is added to streamline get_df's output so each series is in a single column and end date is set as the index.


stephenbernardi:
tested with the following code:
`import os

API_HOST     = 'api.gro-intelligence.com'
client       = GroClient(API_HOST, ACCESS_TOKEN)

brazil_sugar_prod = client.get_data_series(**{'metric_id': 860032, 
                                              'item_id': 13615, 
                                              'region_id': 1029, 
                                              'partner_region_id': 0, 
                                              'frequency_id': 9})
 

# guangdong cash prices soy meal
x = client.GDH('15851987-287-13322-0-1-128')

for s in brazil_sugar_prod:
    client.add_single_data_series(s)

df = client.get_df(index_by_series=True)
df2 = client.get_df(index_by_end_date=True)
`

here is an example of index_by_series output: 
image

here is an example of index_by_end_date output:
image

api/client/gro_client.py
```diff
@@ -1082,9 +1111,16 @@ def GDH(self, gdh_selection, **optional_selections):
         entity_ids = [int(x) for x in gdh_selection.split("-")]
         selection = zip_selections(entity_ids, **optional_selections)
 
-        self.add_single_data_series(selection)
+        gdh_series = self.get_data_series(**selection)
+        if (len(gdh_series) > 1):
+            self._logger.warn("Multiple series returned for GDH selection")
+            return pandas.DataFrame()
+        for series in gdh_series: self.add_single_data_series(series)
+        entity_names = [series[entity_id.replace('id', 'name')] \
```
itsjohnward: There's a couple failing unit tests on this PR.

One of them's related to this line, breaking GDH:

```py
In [4]: client.GDH('860032-274-1215-0-2-9')                                                         
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-4-6dcbd49e07c8> in <module>
----> 1 client.GDH('860032-274-1215-0-2-9')

~/api-client/api/client/gro_client.py in GDH(self, gdh_selection, **optional_selections)
   1118         for series in gdh_series: self.add_single_data_series(series)
   1119         entity_names = [series[entity_id.replace('id', 'name')] \
-> 1120                         for entity_id in DATA_SERIES_UNIQUE_TYPES_ID]
   1121 
   1122         try:

~/api-client/api/client/gro_client.py in <listcomp>(.0)
   1118         for series in gdh_series: self.add_single_data_series(series)
   1119         entity_names = [series[entity_id.replace('id', 'name')] \
-> 1120                         for entity_id in DATA_SERIES_UNIQUE_TYPES_ID]
   1121 
   1122         try:

NameError: free variable 'series' referenced before assignment in enclosing scope
```

itsjohnward (changes requested):

api/client/gro_client.py
```diff
@@ -874,12 +876,39 @@ def get_df(self, show_revisions=False, index_by_series=False):
...
+        if (index_by_series or index_by_end_date) and not self._data_frame.empty:
             columns = intersect(DATA_SERIES_UNIQUE_TYPES_ID, self._data_frame.columns)
+            column_names = [c.replace('id', 'name') for c in columns]
             if len(columns) > 0:
-                indexed_df = self._data_frame.set_index(columns)
-                indexed_df.index.set_names(DATA_SERIES_UNIQUE_TYPES_ID, inplace=True)
+                
+                # add entity names to dataframe
+                series_df = pandas.DataFrame([dict(s) for s in self.get_data_series_list()])
+                df_with_names = self._data_frame.merge(series_df[columns + column_names],
```
nemozen: add copy=False. merge makes a copy by default, so allowing two copies to exist in memory t the same time would cause  problems.

api/client/gro_client.py
```diff
@@ -874,12 +876,39 @@ def get_df(self, show_revisions=False, index_by_series=False):
...
+                                                       on=columns)
+                
+                # add unit name
+                unit_lookup = {unit_id:self.lookup('units', unit_id)['name'] \
```
nemozen: space after `:`

api/client/gro_client.py
```diff
@@ -874,12 +876,39 @@ def get_df(self, show_revisions=False, index_by_series=False):
...
+                               for unit_id in list(self._data_frame.unit_id.unique())}
+                df_with_names['unit_name'] = df_with_names.unit_id.apply(lambda x: unit_lookup[x])
+                column_names.append('unit_name')
+                
+                #set entity names as index
+                indexed_df = df_with_names.set_index(column_names)
+                indexed_df.index.set_names(column_names, inplace=True)
```
nemozen: entity names are not unique in general so indexing by names will mix data series. the underlying index should be as before by entity id. and we can just add the names on top MultiIndex.

api/client/gro_client.py
```diff
@@ -1082,9 +1111,16 @@ def GDH(self, gdh_selection, **optional_selections):
         entity_ids = [int(x) for x in gdh_selection.split("-")]
         selection = zip_selections(entity_ids, **optional_selections)
 
-        self.add_single_data_series(selection)
+        gdh_series = self.get_data_series(**selection)
```
nemozen: we don't want to call get_data series here, as this is at the data retrieval "layer"  while get_data_series is exploration layer.  also don't want to change the semantics of this function... it will return a single series because gdh string is specified that way and this is a selection.

all we would need here is to convert `selection` into `entity names` i.e. get a list of (entity_type, entity_id) from selections and  `entity_names = [self.lookup(entity_type, entity_id) ....]`

nemozen: that said, see comment in get_df(),  entity names are not unique in general so we can't really ise them as the index to get series, we need ids and then have names as just an option output.  so this function can remain unchanged.

api/client/gro_client.py
```diff
@@ -866,6 +866,8 @@ def get_df(self, show_revisions=False, index_by_series=False):
             See https://developers.gro-intelligence.com/data-point-definition.html
             If index_by_series is set, the dataframe is indexed by series.
             See https://developers.gro-intelligence.com/data-series-definition.html
+            If index_by_end_date is set, the dataframe is indexed by end date, with the
```
nemozen: let's not combine two features in one PR. if we index by date do it in a separate pr. also in this implementation, if index_by_series is false then index_by_end_date has no effect which is confusing...

nemozen (changes requested):
1) we need to keep the underlying index to be by entity id and use a multindex for names and 2) the date change should be in a separate pr and either work with/without index_by_series, or not be a separate option but implicit in the latter

api/client/gro_client.py
stephenbernardi: done

api/client/gro_client.py
stephenbernardi: reverted the GDH changes

api/client/gro_client.py
stephenbernardi: reverted the index by name change

stephenbernardi:
I will do the end_date indexing feature in a different PR as suggested, so that part has been removed. I reverted the changes to the index so it should still be indexed by ids. And the entity names are now in the dataframe column headers. @nemozen @itsjohnward please take another look. thanks

api/client/gro_client.py
```diff
@@ -877,8 +877,32 @@ def get_df(self, show_revisions=False, index_by_series=False):
...
+
+                indexed_df = df_with_names.set_index(columns)
                 indexed_df.index.set_names(DATA_SERIES_UNIQUE_TYPES_ID, inplace=True)
+
+                # move entity names to column headers
+                indexed_cols = pandas.MultiIndex.from_frame(indexed_df[name_columns])
+                indexed_df.drop(name_columns, axis=1, inplace=True)
+
+                # indexed_df = pandas.concat([indexed_df], keys=indexed_cols, axis=1)
+                # indexed_df.columns.names = name_columns + ['entities']
```
itsjohnward: I commented out these two lines because they were causing tests to fail. Without them, it seems to work, but I'm still trying to understand what the intent here is

itsjohnward (changes requested):

api/client/gro_client.py
stephenbernardi: that's where the entity names are added to the column headers. removing these lines defeats the purpose of this PR, or we need to find a different way to accomplish it

itsjohnward:
@stephenbernardi I made some changes, so now:

1. `get_df()` has an option to add names as columns, but the index remains unchanged
2. `GDH()` uses names as its index instead of ids

Does that accomplish what you were after here, or did I miss something?

stephenbernardi:
> @stephenbernardi I made some changes, so now:
> 
> 1. `get_df()` has an option to add names as columns, but the index remains unchanged
> 2. `GDH()` uses names as its index instead of ids
> 
> Does that accomplish what you were after here, or did I miss something?

@itsjohnward thanks! This looks pretty close. I think get_df has the desired output. But the index_by_series option can be independent of the include_names option. I don't think we always need to return the names when indexing by series.

stephenbernardi:
also, we should include unit name in the output I think

itsjohnward (approved):

jli:
Hey @stephenbernardi, it looks like JohnW had approved these changes, but this wasn't merged.

Is this still something we'd like to do? I see the associated Jira is still open.

stephenbernardi:
> Hey @stephenbernardi, it looks like JohnW had approved these changes, but this wasn't merged.
> 
> Is this still something we'd like to do? I see the associated Jira is still open.

@jli yes, I still think these changes would be useful. I think @nemozen needs to review before merging

nemozen:
> > Hey @stephenbernardi, it looks like JohnW had approved these changes, but this wasn't merged.
> > Is this still something we'd like to do? I see the associated Jira is still open.
> 
> @jli yes, I still think these changes would be useful. I think @nemozen needs to review before merging

yes had requested some changes back in july .. will take another pass shortly

jli:
Just a thought: should get_data_series also have this option? If so, should we try to implement this for both methods in a generic way?

groclient/client.py
```diff
@@ -874,13 +877,24 @@ def get_df(self, show_revisions=False, index_by_series=False):
...
             )
-        if index_by_series and not self._data_frame.empty:
-            columns = intersect(DATA_SERIES_UNIQUE_TYPES_ID, self._data_frame.columns)
-            if len(columns) > 0:
-                indexed_df = self._data_frame.set_index(columns)
-                indexed_df.index.set_names(DATA_SERIES_UNIQUE_TYPES_ID, inplace=True)
-                return indexed_df.sort_index()
-        return self._data_frame
+
+        df = self._data_frame.copy()
```
nemozen: the data frame can get pretty large so if we make a copy, the client may run out of memory.. How about we just do the name based index in place.

nemozen (commented):
looks good, just one comment re: memory usage

nemozen:
> Just a thought: should get_data_series also have this option? If so, should we try to implement this for both methods in a generic way?

get_data_series already includes names in the response https://developers.gro-intelligence.com/api.html#groclient.GroClient.get_data_series

groclient/client.py
```diff
@@ -850,7 +850,10 @@ def get_top(self, entity_type, num_results=5, **selection):
             self.access_token, self.api_host, entity_type, num_results, **selection
         )
 
-    def get_df(self, show_revisions=False, index_by_series=False):
+    def get_df(self,
+               show_revisions=False,
+               include_names=False,
```
jli: It would be a bit safer to add `include_names` as the last argument. Adding it here could break existing code.

For example, the meaning of `client.get_df(True, True)` would change from setting index_by_series=True to setting index_by_series=False and include_names=True.

jli: Also, could you update the docstring with a brief description of what include_names does (around line 871)?

groclient/client.py
nemozen: fixed

groclient/client.py
nemozen: done

nemozen: done

nemozen (approved):

nemozen:
Updated to not grow memory usage by making a copy with each get_df().

Note that include_names makes the df use 5x more memory (see
https://github.com/gro-intelligence/api-client/pull/236/commits/86c65fcc8ab144f0f94e428213d4689eb2401599 ) so it should be used with caution.

PR #238, opened 2020-07-06
CLEWS-29521: Update metric and item ids for cash_crops.py

itsjohnward:
Vegetation, ETA, and Land Cover have had some migrations since this script was written. Land cover was being automatically redirected, but it looks like the ETA and Vegetation items were deleted, resulting in a 404 error. This PR updates the ids so there are no more deprecated references.

Now cash_crops.py runs.


eliaschneider (approved):
LGTM

PR #239, opened 2020-07-06
CLEWS-29523: pandas dot notation does not add a column

itsjohnward:
Was changed from bracket notation to dot notation in https://github.com/gro-intelligence/api-client/commit/5b0171d836874a5ecf33c5f3b4c9d4fdc7370a04 to be consistent with the date columns' syntax, but pandas does not allow that syntax for adding columns. See https://pandas.pydata.org/docs/user_guide/indexing.html?highlight=attribute%20access#attribute-access

Tested by running brazil_soybeans.ipynb

The existing unit tests for `index_by_series=True` would have also caught this, except that the mock for `mock_data_points` included `source_id` whereas the real `get_data_points()` endpoint does not output source_id. Adjusted the mock so it is more representative of the current output.

### DIFF
--- a/api/client/gro_client.py
+++ b/api/client/gro_client.py
@@ -907,7 +907,7 @@ class GroClient(object):
         # get_data_points response doesn't include the
         # source_id. We add it as a column, in case we have
         # several selections series which differ only by source id.
-        tmp.source_id = data_series["source_id"]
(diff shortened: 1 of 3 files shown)


eliaschneider (approved):
LGTM

PR #240, opened 2020-07-06
CLEWS-29522: Add "(daytime)" to land temperature search in Brazil Sao Paulo Sugarcane Crop Model

itsjohnward:
`model.search_for_entity('items', "Land temperature")` was matching on

```py
{'aliases': ['Land temperature (daytime) masked by Cropland'],
 'contains': [],
 'definition': 'Daytime land surface temperature value for a given period, masked by cropland.',
 'id': 16497,
 'name': 'Land temperature (daytime) on cropland',
 'belongsTo': [10089]}
```

So the notebook was getting no data for LST. By appending " (daytime)" to the search term it now matches on the expected data series.

Another issue discovered while working on this, however, is that it's assuming there is TRMM data for the latest year, but TRMM stops at the end of 2019. I have not attempted to resolve that and only fixed the search term issue.


itsjohnward:
Declined in favor of @biquanlin's CLEWS-27620 PR

PR #241, opened 2020-07-06
Fall back to None status code if APIError response is not an HTTP response

itsjohnward:
To avoid the error

```py
AttributeError: 'ConnectionError' object has no attribute 'status_code'
```

Now, when a ConnectionError is raised, APIError parses it correctly:

```py
APIError: (ConnectionError(), 5, 'https://api.gro-intelligence.com/v2/regions', {'ids': [1215]})
```


eliaschneider (approved):
LGTM

PR #249, opened 2020-08-18
CLEWS-29688 (refactor only): Move gro_client cli out of client.py

itsjohnward:
* move gro_client command line interface behavior out of `groclient/client.py` and into `groclient/cli.py` instead.
* Add a `__main__.py` [module](https://docs.python.org/3/library/__main__.html), which invokes the cli. So now you can use either `gro_client`, or `python -m groclient` to execute the cli.
* remove the undocumented, deprecated extra methods in client.py. Note, they were not included in the aliasing for the name change anyway https://github.com/gro-intelligence/api-client/blob/development/api/client/gro_client/__init__.py, so they've already been removed from the `api.client.gro_client` namespace.

This PR should include no functional change. Just moving code from one file to another.


ballabusta (approved):
Refactoring looks good to me

PR #252, opened 2020-08-24
CLEWS-31143 Avoid unhashable dicts when adding data series

itsjohnward:
Related to https://github.com/gro-intelligence/api-client/pull/147 where metadata is unhashable.

So, if using `get_data_series()` and passing the result directly into `add_single_data_series()`, it would fail if the data series had any metadata. Instead, run the data series through `unnest_dict`, so if the input is:

```py
{
        "metric_id": 860032,  # TODO: add names
        "item_id": 274,
        "region_id": 1215,
        "partner_region_id": 0,
        "frequency_id": 9,
        "source_id": 2,
        "metadata": {"includes_historical_region": True}
}
```

it gets converted to

```py
{
        "metric_id": 860032,  # TODO: add names
        "item_id": 274,
        "region_id": 1215,
        "partner_region_id": 0,
        "frequency_id": 9,
        "source_id": 2,
        "metadata_includes_historical_region": True}
}
```

which then *is* hashable and generates a frozen set like:

```py
frozenset({('partner_region_id', 0), ('metric_id', 860032), ('source_id', 2), ('region_id', 1215), ('item_id', 274), ('metadata_includes_historical_region', True), ('frequency_id', 9)})
```

This is preferable over filtering out data series properties that don't appear in `DATA_SERIES_UNIQUE_TYPES_ID` because `add_single_data_series` should support extra options like `at_time` and `show_revisions`.

### DIFF
--- a/groclient/client.py
+++ b/groclient/client.py
@@ -1,6 +1,4 @@
 from __future__ import print_function
-from builtins import str
-from random import random
 import functools
 import itertools
 import time
@@ -20,7 +18,6 @@ from groclient.utils import intersect, zip_selections, dict_unnest
 from groclient.lib import APIError
 
 import pandas
-import unicodecsv
 from tornado import gen
 from tornado.escape import json_decode
 from tornado.httpclient import AsyncHTTPClient, HTTPRequest, HTTPError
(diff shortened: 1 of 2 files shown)


eliaschneider (approved):
LGTM

seungatGro (approved):
LGTM. tested unnesting dictionary

PR #257, opened 2020-09-30
feat: teardown tornado resources when groclient is deleted

itsjohnward:
My test script:

```py
from tqdm import tqdm

import os

regions = {}

counter = 0
for region in tqdm(client.get_descendant_regions(0, include_details=False)):
    try:
        regions[region['id']] = client.lookup('regions', region['id'])['name']
        counter += 1
    except Exception as e:
        print(counter, e)
        break
```

Before this change, after 80 GroClients get created, there's an error:

```py
HTTPSConnectionPool(host='api.gro-intelligence.com', port=443): Max retries exceeded with url: /v2/regions?ids=1073 (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x11931c8d0>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))` error
```

But, with this `__del__` method in place, I'm able to initialize more. Tested up to 350.

If I modify my test script to keep all the created clients in a list, then the limit of 80 *still exists*:

```py
from tqdm import tqdm

import os

clients = []
regions = {}

counter = 0
for region in tqdm(client.get_descendant_regions(0, include_details=False)):
    try:
        clients.append(GroClient('api.gro-intelligence.com', os.environ['GROAPI_TOKEN']))
        regions[region['id']] = clients[-1].lookup('regions', region['id'])['name']
        counter += 1
    except Exception as e:
        print(counter, e)
        break
```

But if the variable's being overwritten, or if `del` is explicitly called, then it's good.

### DIFF
--- a/groclient/client.py
+++ b/groclient/client.py
@@ -73,6 +73,11 @@ class GroClient(object):
             )
             self._async_http_client = None
             self._ioloop = None
+        
+    def __del__(self):
+        self._async_http_client.close()
+        self._ioloop.stop()
+        self._ioloop.close()
 
     def get_logger(self):
         return self._logger


iba3:
With this change, up to 350 clients can be spun up in one session or has it only been tested up to 350 clients and possibly more can be created?

This is a good change. Although it seems to me on the user side, if the example or something similar is what's happening in practice it probably is sliding into behavior that probably shouldn't be encouraged. As the user should likely just use the same client (or a couple in order to parallelize further) to download subregional data.

iba3 (approved):

itsjohnward:
> With this change, up to 350 clients can be spun up in one session or has it only been tested up to 350 clients and possibly more can be created?

The latter. I just cancelled out after it got to 350, but it probably could go much higher.

> This is a good change. Although it seems to me on the user side, if the example or something similar is what's happening in practice it probably is sliding into behavior that probably shouldn't be encouraged. As the user should likely just use the same client (or a couple in order to parallelize further) to download subregional data.

Exactly. Re-using a single client would certainly be a better usage of resources than creating/destroying clients whenever you want to request something. But, if you did want to do the inefficient thing for some reason, without this change the error you run into is not clear at all that this is what's causing it. Just allowing you to do it if you want to is probably easier than explaining the problem.

Imagine if you re-ran a Jupyter cell that creates a client over and over and then inexplicably after 80 times it stopped working until you restarted your kernel. It's just weird behavior.

jli:
Just FYI, it seems like the root cause of this problem is that `tornado.ioloop.IOLoop` opens some file descriptors, and after creating ~80 IOLoops, subsequent network requests fail due to too many open file descriptors (the OS X default limit seems to be ~256):

```shell
$ python -c '[open("/tmp/"+str(i),"w") for i in range(255)]'
  File "<string>", line 1, in <module>
IOError: [Errno 24] Too many open files: '/tmp/253'
```

PR #261, opened 2020-10-16
notebook for Drought and El Nino/La Nina

nemozen:

PR #269, opened 2020-10-21
CLEWS-32963 fix doc gen UnicodeDecodeError on Shippable

jli:
sphinxcontrib-versioning claims to only support up to Python 3.5, but
manual testing on Python 3.7 shows that it works. (It hasn't been
updated since 2016, so that's probably why it doesn't list support for
recent versions.)

This should fix the nondeterministic-seeming documentation generation
errors we see on Shippable, which has been blocking PR merges.
CLEWS-32963

### DIFF
--- a/shippable.yml
+++ b/shippable.yml
@@ -38,8 +38,14 @@ build:
     # https://github.com/sphinx-contrib/sphinxcontrib-versioning does not support the latest 3.x
     # versions, which we do want to run unit tests on. That is why 2.x is used, rather than adding
     # an older 3.x version to the build matrix just to support docs.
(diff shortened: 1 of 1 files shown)


muzigao (approved):

jli:
FYI: I dug into why sphinx-versioning was failing with Python 2. It's because PR #261 introduced a file with a Unicode character (a new ipython notebook: `El Niño, La Niña and Droughts in East Africa.ipynb`).

So, I guess sphinxcontrib-versioning clones the whole repo, including branches, and processes files even in other branches, for some reason...

PR #253, opened 2020-08-27
CLEWS-23863: remove duplicates from rank_series_by_source

itsjohnward:
Follow-up to https://github.com/gro-intelligence/api-client/pull/250#issuecomment-677676097

Currently, if you pass `rank_series_by_source` a list of data series, as output from `get_data_series`, the output can be confusing since it's duplicating those input series and just changing source ids. So source names and start/end dates and estimated data counts can all be misrepresented. For example:

```py
data_series_list = client.get_data_series(metric_id=570001, item_id=95, region_id=1094, frequency_id=9)
for data_series in data_series_list:
    print(data_series['source_id'], data_series['start_date'], data_series['end_date'], data_series['data_count_estimate'])
```

this shows the available data series as:

```py
2 1961-01-01T00:00:00.000Z 2018-12-31T00:00:00.000Z 59
14 1960-01-01T00:00:00.000Z 2021-03-31T00:00:00.000Z 62
19 1999-01-01T00:00:00.000Z 2016-12-31T00:00:00.000Z 19
50 1997-07-01T00:00:00.000Z 2018-06-30T00:00:00.000Z 22
```

On the development branch, the output of

```py
for data_series in client.rank_series_by_source(data_series_list):
    print(data_series['source_id'], data_series['start_date'], data_series['end_date'], data_series['data_count_estimate'])
```

is

```py
50 1961-01-01T00:00:00.000Z 2018-12-31T00:00:00.000Z 59
14 1961-01-01T00:00:00.000Z 2018-12-31T00:00:00.000Z 59
2 1961-01-01T00:00:00.000Z 2018-12-31T00:00:00.000Z 59
19 1961-01-01T00:00:00.000Z 2018-12-31T00:00:00.000Z 59
50 1960-01-01T00:00:00.000Z 2021-03-31T00:00:00.000Z 62
14 1960-01-01T00:00:00.000Z 2021-03-31T00:00:00.000Z 62
2 1960-01-01T00:00:00.000Z 2021-03-31T00:00:00.000Z 62
19 1960-01-01T00:00:00.000Z 2021-03-31T00:00:00.000Z 62
50 1999-01-01T00:00:00.000Z 2016-12-31T00:00:00.000Z 19
14 1999-01-01T00:00:00.000Z 2016-12-31T00:00:00.000Z 19
2 1999-01-01T00:00:00.000Z 2016-12-31T00:00:00.000Z 19
19 1999-01-01T00:00:00.000Z 2016-12-31T00:00:00.000Z 19
[...]

PR #270, opened 2020-10-22
CLEWS-33028: support list of ids in rank_series_by_source

muzigao:
**Issue**
Right now when passing queries with list of ids into rank_series_by_source(), it returns empty []
It's throwing `ValueError` from https://github.com/gro-intelligence/api-client/blob/development/groclient/lib.py#L426

The bug was introduced in https://github.com/gro-intelligence/api-client/pull/253/files 
that implementation only handles the case when selection includes single id.

**Solution**
parse series key with ast.literal_eval(), which should handle both single id and list of ids

**Testing**
Tested locally with Raphael's code, it should return ranked series now
```

region_id = 1215
item_id = 274
metric_id = 570001
regions = [
    r['id'] for r in client.get_descendant_regions(region_id, 4, include_historical=False, include_details=False)]

query = [{
    "metric_id": metric_id,
    "region_id": regions,
    "item_id": item_id, }]

res = list(client.rank_series_by_source(query))
print(res)
```

```
{'metric_id': 570001, 'item_id': 274, 'region_id': [13051, 13052, 13053, 13054, 13055, 13056, 13057, 13058, 13059, 13060, 13061, 13062, 13063, 13064, 13065, 13066, 13067, 13068, 13069, 13070, 13071, 13072, 13073, 13074, 13075, 13076, 13077, 13078, 13079, 13080, 13081, 13082, 13083, 13084, 13085, 13086, 13087, 13088, 13089, 13090, 13091, 13092, 13093, 13094, 13095, 13096, 13097, 13098, 13099, 13100, 13101], 'source_id': 25}]
```

### DIFF
--- a/groclient/lib.py
+++ b/groclient/lib.py
@@ -15,7 +15,6 @@ import logging
 import requests
 import time
 import platform
-import ast
 from pkg_resources import get_distribution, DistributionNotFound
 try:
     # functools are native in Python 3.2.3+
@@ -423,7 +422,7 @@ def rank_series_by_source(access_token, api_host, selections_list):
 
     for series_key, series_by_source_id in series_map.items():
         series_without_source = {
-            type_id: ast.literal_eval(series_key.split('.')[idx])
+            type_id: json.loads(series_key.split('.')[idx])
             for idx, type_id in enumerate(DATA_SERIES_UNIQUE_TYPES_ID)
             if type_id != 'source_id' and series_key.split('.')[idx] != 'None'
         }
--- a/groclient/lib.py
+++ b/groclient/lib.py
@@ -15,6 +15,7 @@ import logging
 import requests
 import time
 import platform
+import ast
 from pkg_resources import get_distribution, DistributionNotFound
 try:
     # functools are native in Python 3.2.3+
@@ -423,7 +424,7 @@ def rank_series_by_source(access_token, api_host, selections_list):
     for series_key, series_by_source_id in series_map.items():
         try:
             series_without_source = {
-                type_id: int(series_key.split('.')[idx])
+                type_id: ast.literal_eval(series_key.split('.')[idx])
                 for idx, type_id in enumerate(DATA_SERIES_UNIQUE_TYPES_ID)
                 if type_id != 'source_id' and series_key.split('.')[idx] != 'None'
             }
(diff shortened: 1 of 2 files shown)


groclient/lib.py
```diff
@@ -423,7 +424,7 @@ def rank_series_by_source(access_token, api_host, selections_list):
     for series_key, series_by_source_id in series_map.items():
         try:
             series_without_source = {
-                type_id: int(series_key.split('.')[idx])
+                type_id: ast.literal_eval(series_key.split('.')[idx])
                 for idx, type_id in enumerate(DATA_SERIES_UNIQUE_TYPES_ID)
                 if type_id != 'source_id' and series_key.split('.')[idx] != 'None'
             }
```
jli: Is it necessary for this piece of code to be within the `try`? Perhaps we can move it outside? Based on the `# empty response` comment, my guess is that the `try` is meant to catch errors from `get_source_ranking`.

If we move it out, we probably would have caught this bug via tests.

cn1036 (approved):

eliaschneider (approved):
LGTM

groclient/lib.py
muzigao: That's a good point - I just removed it completely. i feel it's just misleading to return empty result silently

ballabusta (approved):

groclient/lib.py
```diff
@@ -421,17 +421,15 @@ def rank_series_by_source(access_token, api_host, selections_list):
...
-                for idx, type_id in enumerate(DATA_SERIES_UNIQUE_TYPES_ID)
-                if type_id != 'source_id' and series_key.split('.')[idx] != 'None'
-            }
-            source_ids = get_source_ranking(access_token,
-                                            api_host,
-                                            series_without_source)
-        except ValueError:
-            continue  # empty response
+        series_without_source = {
+            type_id: json.loads(series_key.split('.')[idx])
```
jli, line 425: optional/nit: maybe do `json.dumps()` instead of `str()` on line 414 to mirror the json.dumps here?

jli (approved):

groclient/lib.py
muzigao: good idea. change pushed

PR #271, opened 2020-10-23
fix deprecated usage of `warn`

jli:
Previously, the code was generating a DeprecationWarnings about our
usage of logging.Logger.warn, for example when running pytest:

======================== warnings summary ========================
groclient/client_test.py::GroClientTests::test_GDH
  /Users/johnli/src/api-client/groclient/client.py:1083: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
    self._logger.warn("GDH returned no data")

-- Docs: https://docs.pytest.org/en/stable/warnings.html
======================== 63 passed, 1 warning in 4.24s ========================

### DIFF
--- a/groclient/client.py
+++ b/groclient/client.py
@@ -68,12 +68,12 @@ class GroClient(object):
             self._async_http_client = AsyncHTTPClient()
             self._ioloop = IOLoop()
         except Exception as e:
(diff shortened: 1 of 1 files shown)


muzigao (approved):
LGTM

PR #274, opened 2020-10-27
add generic get_descendant function

eliaschneider:
Added a more function which returns all child included in an entity_id. 
The function _get_descendant_regions_ works specifically only for regions, _get_descendant_ work also for items and metrics

Script to test it: 
````
import os

item_id = list(client.search_and_lookup('items', 'Fruits'))[0]
print('name: {}, id: {}'.format(item_id['name'], item_id['id']))
res = client.get_descendant('items', item_id['id'], include_details=True, distance=1)
print('Directly include in {}: '.format(item_id['name']) + ' ,'.join([item['name'] for item in res]))
res = client.get_descendant('items', item_id['id'], include_details=False)
print('Total items included in {}: {}'.format(item_id['name'], len(res)))

metric_id = list(client.search_and_lookup('metrics', 'NDVI'))[0]
print('name: {}, id: {}'.format(metric_id['name'], metric_id['id']))
res = client.get_descendant('metrics', metric_id['id'], include_details=True, distance=1)
print('Directly include in {}: '.format(metric_id['name']) + ' ,'.join([metr['name'] for metr in res]))
res = client.get_descendant('metrics', metric_id['id'], include_details=False)
print('Total items included in {}: {}'.format(metric_id['name'], len(res)))


output: 


name: Fruits, id: 18063
Directly include in Fruits: Fruit, dried, NES ,Fruit, fresh, NES ,Tropical fruit, fresh, NES ,Fruits, NES ,Fruit crops ,Tropical and subtropical fruit ,Tropical and subtropical fruit, NES ,Temperate fruits ,Produce, packaged
Total items included in Fruits: 2871
name: Vegetation, id: 103
Directly include in Vegetation: Vegetation Indices ,Vegetation Anomalies ,Biomass Burned ,Burned Area
Total items included in Vegetation: 10
```

### DIFF
--- a/groclient/client.py
+++ b/groclient/client.py
@@ -705,7 +705,11 @@ class GroClient(object):
         distance=None,
         include_details=True,
     ):
-        """Look up details of all entities level contained by an item/metric/regions.
+        """Given an item/metric/region id return all ids contained in the graph, and their relative details
+
+        Similar to :meth:~.get_descendant_regions, but also works on items and metrics. This method has 
+        a distance parameter (which returns all nested child entities) instead of a descendant_level 
+        parameter (which only returns child entities at a given depth/level).
 
         Parameters
         ----------
@@ -728,11 +732,11 @@ class GroClient(object):
                 [{
                     'id': 134,
                     'name': 'Cattle hides, wet-salted',
-                    'definition': 'Hides and skins of domesticated cattle—animals ...',
+                    'definition': 'Hides and skins of domesticated cattle-animals ...',
                 } , {
                     'id': 382,
                     'name': 'Calf skins, wet-salted',
-                    'definition': 'Wet-salted hides and skins of calves—animals of ...'
+                    'definition': 'Wet-salted hides and skins of calves-animals of ...'
                 }, ...]
 
             See output of :meth:`~.lookup`
--- a/groclient/client.py
+++ b/groclient/client.py
@@ -707,8 +707,6 @@ class GroClient(object):
     ):
         """Look up details of all entities level contained by an item/metric/regions.
 
-        Given any region by id, get all the descendant regions that are of the specified level.
-
         Parameters
         ----------
         entity_type : { 'metrics', 'items', 'regions' }
--- a/groclient/client.py
+++ b/groclient/client.py
@@ -705,7 +705,8 @@ class GroClient(object):
         distance=None,
         include_details=True,
     ):
-        """Given an item/metric/region id return all ids contained in the graph, and their relative details
+        """Given an item, metric or region, returns all its descendants i.e. 
+        entities that are "contained" in the given entity
 
         Similar to :meth:~.get_descendant_regions, but also works on items and metrics. This method has 
         a distance parameter (which returns all nested child entities) instead of a descendant_level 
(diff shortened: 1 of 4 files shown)


groclient/client.py
```diff
@@ -698,6 +698,58 @@ def get_geojson(self, region_id, zoom_level=7):
...
+    def get_descendant(
+        self,
+        entity_type,
+        entity_id,
+        distance=None,
+        include_details=True,
+    ):
+        """Look up details of all entities level contained by an item/metric/regions.
+
+        Given any region by id, get all the descendant regions that are of the specified level.
```
jli: It seems like this line in the docstring doesn't apply to this method?

jli (commented):
Should we implement `get_descendant_regions` via `get_descendant`?
Or, perhaps `get_descendant` should only work on items and metrics? (I see that `get_descendant_regions` has another option, `include_historical`.)

Can we add tests? :)

groclient/client.py
eliaschneider: Fix it

eliaschneider:
@jli We could refactor  `get_descendant_regions` to make it a special case of `get_descendant` but we also have another option that is the possibility to select `level` other than filter out `historical` regions.

groclient/client.py
```diff
@@ -698,6 +698,56 @@ def get_geojson(self, region_id, zoom_level=7):
...
         return lib.get_geojson(self.access_token, self.api_host, region_id, zoom_level)
 
+    def get_descendant(
+        self,
+        entity_type,
+        entity_id,
+        distance=None,
+        include_details=True,
+    ):
+        """Look up details of all entities level contained by an item/metric/regions.
```
jli: I'm not sure I'm understanding the docstring - what does "level" mean? Also, I think it would be nice for users to specify how this differs from get_descendant_regions. Perhaps something like this?

> Look up details of entities contained by an item/metric/region.
>
> Similar to :meth:`~.get_descendant_regions`, but also works on items and metrics. This method has a `distance` parameter (which returns all nested child entities) instead of a `descendant_level` parameter (which only returns child entities at a given depth/level).

We could also modify the docstring for `get_descendant_regions` to just say something like:
> Also see :meth:`~.get_descendant`, which works on items and metrics as well.

jli:
Hm, also I see Shippable is failing for Python 2:

> E   SyntaxError: Non-ASCII character '\xe2' in file /root/src/github.com/gro-intelligence/api-client/groclient/client.py on line 732, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

Weird... Can you check whether you accidentally included some UTF-8 text in your file..?

jli (approved):

groclient/client.py
```diff
@@ -698,6 +698,60 @@ def get_geojson(self, region_id, zoom_level=7):
...
         return lib.get_geojson(self.access_token, self.api_host, region_id, zoom_level)
 
+    def get_descendant(
+        self,
+        entity_type,
+        entity_id,
+        distance=None,
+        include_details=True,
+    ):
+        """Given an item/metric/region id return all ids contained in the graph, and their relative details
```
nemozen: "all ids contained in the graph" is incorrect, sounds like we're returning the entire graph.  How about "given an item, metric or region, returns all its descendants i.e. entities that are "contained" in the given entity".

muzigao (approved):
looks good to me!

PR #279, opened 2020-11-04
CLEWS-33186 conf_interval was not converted into requested unit

cn1036:


cn1036:
Issue is reported as https://grointelligence.atlassian.net/browse/CLEWS-33186
Added unit conversion for confidence interval if exists

groclient/lib.py
```diff
@@ -150,6 +150,16 @@ def get_version_info():
     return _VERSIONS.copy()
 
 
+def compute_value(value, from_convert_factor, to_convert_factor):
```
muzigao: minor - might be better to call it `convert_value` , "compute" is little unclear what's the calculation is doing

groclient/client.py
```diff
@@ -1445,12 +1445,12 @@ def convert_unit(self, point, target_unit_id):
...
+
         if point.get("value") is not None:
-            value_in_base_unit = (
-                point["value"] * from_convert_factor.get("factor")
-            ) + from_convert_factor.get("offset", 0)
-            point["value"] = float(
-                value_in_base_unit - to_convert_factor.get("offset", 0)
-            ) / to_convert_factor.get("factor")
+            point["value"] = lib.compute_value(point["value"], from_convert_factor, to_convert_factor)
+        if point.get("metadata") is not None:
```
tolothy: could we add a unit test case covering this condition?

joysaikat:
Will it be practical to have a unit id in the `metadata` dictionary as well? Although, when this is merged it will be redundant; it would have been more enlightening before this PR.

groclient/lib.py
joysaikat: If the goal of this method is to convert the unit, could we have used `client.convert_unit`?

tolothy:
> Will it be practical to have a unit id in the `metadata` dictionary as well? Although, when this is merged it will be redundant; it would have been more enlightening before this PR.

@joysaikat I think that is overall a good practice going forward. But to deal with existing entries in fact tables, we have to assume that the unit_id, if applicable, is the same as the one specified in column `input_unit_id`

cn1036:
@tolothy @joysaikat 
When converting the value of conf_interval, we do assume that both data point and its confidence interval are reported in the same unit(input_unit_id). Besides the api client code, this assumptions applies to webapp(balance sheet) as well. So, if this is desired, we can add unit_id to metadata(or should we call it conf_interval_unit_id since metadata may contain other attributes that are irrelevant to conf interval?).

cn1036:
by the way, do we need to keep both conf_interval and confInterval? confInterval is used by webapp, not sure if it's necessary for api client requests

joysaikat (commented):
> by the way, do we need to keep both conf_interval and confInterval? confInterval is used by webapp, not sure if it's necessary for api client requests

@cn1036 No we do not. @tolothy can speak more on this probably. We should remove confInterval if possible.

groclient/lib.py
cn1036: this is a helper function which deals with the calculation, while client.convert_unit focuses on data point level.

cn1036: renamed

groclient/client.py
cn1036: added

groclient/client_test.py
```diff
@@ -479,6 +479,14 @@ def test_convert_unit(self):
             self.client.convert_unit({"value": 1, "unit_id": 37}, 36),
             {"value": -17.5, "unit_id": 36},
         )
+        self.assertEqual(
+            self.client.convert_unit({"value": 20, "unit_id": 10, "metadata": {"conf_interval": 2}}, 14),
+            {"value": 0.02, "metadata": {"conf_interval": 0.002}, "unit_id": 14},
+        )
+        self.assertEqual(
+            self.client.convert_unit({"value": 20, "unit_id": 10, "metadata": {"conf_interval": None}}, 14),
```
tolothy: I think we barely have the case with `"metadata": {"conf_interval": None}`. If no confidence interval is provided, it would rather leave that attribute out. On the other hand, you might want to add a test case with `"metadata": {}`

tolothy (approved):
lgtm

muzigao (approved):
LGTM

PR #280, opened 2020-11-05
CLEWS-33473 convert metadata.confInterval into snake_case

cn1036:
We are going to remove the redundant attribute 'conf_interval' in v2/data, so that the endpoint is always using camelcase if responseType is set to 'list_of_series'. Updated `lib.list_of_series_to_single_series()` to convert the format of metadata.
https://bitbucket.org/grointelligence/gro/pull-requests/18324/clews-33473-remove-metadataconf_interval/diff

### DIFF
--- a/groclient/lib.py
+++ b/groclient/lib.py
@@ -503,6 +503,9 @@ def list_of_series_to_single_series(series_list, add_belongs_to=False, include_h
                 'frequency_id': series['series'].get('frequencyId', None)
                 # 'source_id': series['series'].get('sourceId', None), TODO: add source to output
             }
+            if formatted_point['metadata'].get('confInterval') is not None:
+                formatted_point['metadata']['conf_interval'] = formatted_point['metadata'].pop('confInterval')
+
             if add_belongs_to:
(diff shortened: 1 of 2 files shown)


groclient/lib.py
```diff
@@ -503,6 +503,9 @@ def list_of_series_to_single_series(series_list, add_belongs_to=False, include_h
                 'frequency_id': series['series'].get('frequencyId', None)
                 # 'source_id': series['series'].get('sourceId', None), TODO: add source to output
             }
+            if formatted_point['metadata'].get('confInterval') is not None:
```
muzigao, line 506: can we also update unit test for this ?

joysaikat:
@cn1036 , I thought we were planning to remove `confInterval` and not `conf_interval`. What are we removing?

cn1036:
@joysaikat sorry for the confusion. updated description

groclient/lib.py
cn1036: done

muzigao (approved):

PR #282, opened 2020-11-16
API Recipes

JGrinst:
Creating a new API Cookbook Section
Adding sub-sections of recipes within the Cookbook
Adding recipe examples.

Creating a PR to review changes, but not ready for review or merge until specified


jli:
I was able to build this branch successfully after updating Sphinx (and replacing the sphinxcontrib-versioning extension). But I ran into another issue: nbsphinx requires the pandoc tool: https://nbsphinx.readthedocs.io/en/0.8.0/installation.html#pandoc

Pandoc [isn't pip-installable](https://pandoc.org/installing.html). This poses a problem for our current approach of building the documentation via Shippable automatically, since the environment that Shippable uses (almost certainly) doesn't have pandoc installed.

Some ideas (sorted by ascending amount of work, roughly):
- Don't include notebooks directly in docs: GitHub does a decent job at rendering notebooks ([example](https://github.com/gro-intelligence/api-client/blob/development/api/client/samples/at-time-query-examples.ipynb)).
- Don't use nbsphinx: there's a `jupyter nbconvert --to rst` command that (should) be able to generate RST. We might be able to automate this.
- Build docs locally: we could give up on automating the documentation generation, and just have people build the docs locally whenever we make a change.
- Docker: I believe Shippable supports custom Docker containers, so we could try to use that to make pandoc available. Would require some figuring out.
- GitHub Actions: seems to [support pandoc](https://github.com/pandoc/pandoc-action-example), so we could try to use this for docs generation instead of Shippable. Not sure if this costs money, would require figuring out, and adds more complexity to our CI setup.

I'll play around with getting `jupyter nbconvert`, since that seems like it would be the best option if it works.

jli:
I was wrong about it being difficult to run pandoc from Shippable: `apt-get install` works just fine! (tested on a separate branch)

Status of this PR is that it's blocked on me updating Sphinx (and migrating from sphinxcontrib-versioning to sphinx-multiversion) so that we can use the nbsphinx plugin.

.circleci/config.yml
```diff
@@ -57,6 +57,9 @@ jobs:
           name: "Ensure latest docs build without warnings or errors"
           command: >-
             poetry install -E docs &&
+            poetry run pip install ipython &&
```
seungatGro, line 60: @kmetallic04 can we move this to pyproject.toml so it is installed with the command above?

seungatGro: tested above, does not work with poetry. this should be good to go if this is the only method to install python on ci

seungatGro (approved):

.circleci/config.yml
kmetallic04: Hmmm.... I think there's a version of `ipython` (version `7.16.1`) that's compatible with the current requirements for python within poetry (`>=3.6.1,<4.0.0`). I think we can update `pyproject.toml` and `poetry.lock` instead.

kmetallic04: Done. Added `ipython` as an `extra` for `recipes`. No need for using `pip`.

muzigao:
closing this PR (for similar region issue).  Branch will be kept

PR #283, opened 2020-11-19
update warnings to reflect that api/client/samples is not deprecated

nemozen:
we are keeping the samples accessible outside of the groclient distribution


jli:
Thanks, this was on my TODO list.

My understanding of CLEWS-16749 was that we would eventually get rid of the top-level `api/` directory completely.
What about moving the samples to another top-level directory, like `samples/` or `examples/`? This seems like it makes more sense that having these live within the `api.client` module path.

nemozen:
> Thanks, this was on my TODO list.
> 
> My understanding of CLEWS-16749 was that we would eventually get rid of the top-level `api/` directory completely.
> What about moving the samples to another top-level directory, like `samples/` or `examples/`? This seems like it makes more sense that having these live within the `api.client` module path.

yes it woudl be slightly cleaner to move /api/client/samples to just /samples  but the samples have been widely shared and linked to so better to not break links (for now at least)

jli:
Ok, preserving the existing path for now makes sense.

In terms of moving where the warning logging happens, seems like we should also add it to the other submodules in `api/client`? Namely: batch_client, cfg, and crop_model.

nemozen:
> Ok, preserving the existing path for now makes sense.
> 
> In terms of moving where the warning logging happens, seems like we should also add it to the other submodules in `api/client`? Namely: batch_client, cfg, and crop_model.

done

jli (approved):

jli:
I noticed an issue with this change.

Since `api/client/__init__.py` makes `Client` available, a user can do the following without receiving any errors:
- `from api.client import Client`
- or `import api.client; api.client.Client(...)`

I'm not sure if there's any way to only warn on imports *other* than the samples without breaking backwards compatibility.

PR #285, opened 2020-11-22
add HTTP 502 to response status that trigger retry

nemozen:
separating this out from https://github.com/gro-intelligence/api-client/pull/227


ash-gro:
This (placement of 502 code) is the trivial part. There is a more serious problem which was addressed in the original commit of client.py of the new similar region branch, see
https://github.com/gro-intelligence/api-client/pull/227/commits/d200b3a5741b630618635aa1e6de1d9588160ddc
Note extra break statement added to avoid treating other exceptions as success.

If you want to move updates to client.py into a separate PR (which is probably a good idea),  you might want to move this more important update as well

jli:
@ash-gro it seems like the changes in that commit got lost later? Looking at the overall [files-changed view](https://github.com/gro-intelligence/api-client/pull/227/files?file-filters%5B%5D=.py&hide-deleted-files=true#diff-06f5cb8f2c2d813b8fb4113282468eb0a49815db12b3e5d8395ba0ef0ad81870) for #227, I don't see that extra break in client.py. 

(Also in the commit you linked, the 502 seems to be placed in the "fail immediately" elif conditional)

jli (approved):
this change lgtm, though note ash's comment about ignoring other status codes (also if the exception didn't have the code attribute, the loop will continue)

ash-gro:
There were two changes to client.py made within similar regions (#227).  One was in the very original commit and included 
1. the extra 'break' (which is actually a bug fix)
2. adding 502 code

 The later update (one which @nemozen extracted into this separate PR) just changes where to place particular code 502 and was done at the request of @itsjohnward. It is indeed legitimate, but I want to make sure the more important first one is not lost which seems exactly what has happened somehow (probably due to restructuring of overall API code - relevant code was moved to groclient).

jli:
@ash-gro it seems like we should just remove the `elif` conditional and always `break` when the status code isn't within a list of whitelisted/retryable codes, right?

Can you say how you ran into this issue? By my reading, not adding the `break` just means that we will retry a few extra times before finally raising the exception.

muzigao (approved):

ash-gro:
Yes, removing `elif` should have the same effect but I would guess the 400-level codes were singled out for some reason (maybe some additional TODO processing), so I did not want to remove this

Consider what happens if we get any exception not covered by the explicitly stated status_code's. I guess this is rare but that's what happened to me (and conceptually you can not expect to have exhaustive set of error codes there). We are catching all exceptions, and in the original code (no extra break) we simply do nothing and fall through to the end of `except Exception as e:` clause. At this point exception is considered handled and we proceed to the "Request was successful" part which is clearly not what actually happened. This leads to returning some unknown results which caller sees as valid.

nemozen:
> Yes, removing `elif` should have the same effect but I would guess the 400-level codes were singled out for some reason (maybe some additional TODO processing), so I did not want to remove this
> 
> Consider what happens if we get any exception not covered by the explicitly stated status_code's. I guess this is rare but that's what happened to me (and conceptually you can not expect to have exhaustive set of error codes there). We are catching all exceptions, and in the original code (no extra break) we simply do nothing and fall through to the end of `except Exception as e:` clause. At this point exception is considered handled and we proceed to the "Request was successful" part which is clearly not what actually happened. This leads to returning some unknown results which caller sees as valid.

yes the api server tries to respect the standard meaning of http response codes.

- those 4xx status codes  => "not found",  i.e. the api server is saying there is no error, simply the data does not exist (400 means malformed request, 401 is permission denied, 402 is when the client has reached their data limit and 404 is when the entity or series does not exist) , so we don't want to retry in those cases since the answer will be the same

-  429 => "too many requests",  https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429 so for that one we do want to retry but more slowly hence the exponential backoff

-  5xx   => server processing failures where do we want to retry.

and when it is none of the above, we could arguably be more conservative and bail out. But since we already limit it to  MAX_RETRIES, it's better to give the benefit of the doubt and try again.

jli:
@ash-gro I think I ran into the scenario you were describing with the client silently ignoring errors.

```python
import os

# Create a new GroClient (which reuses AsyncHTTPClient), then cause the
# original GroClient's destructor to run (which closes the shared
# AsyncHTTPClient).

# Logs `fetch() called on closed AsyncHTTPClient`. data is just [None]
data = client.batch_async_get_data_points([
    {'metric_id': 860032, 'item_id': 274, 'region_id': 1215, 'frequency_id': 9, 'source_id': 2},
])
print(data)
```

In this test script, the `batch_async_get_data_points` call triggers a bug in GroClient (see #297), the error is logged but "handled", and `data` is just `[None]`.

When I add the `break` that you mention, the resulting data becomes `[BatchError(RuntimeError('fetch() called on closed AsyncHTTPClient'), ...)]`.

Does that match what you were trying to achieve with your original commit?

ash-gro:
Certainly feels like the same issue (silently ignoring unlisted errors and returning some output as if no error occurred) ...

PR #288, opened 2020-11-25
disabling Shippable on gh-pages step 1

jli:
Shippable keeps running (and failing) on the gh-pages branch
unnecessarily. We need to add a shippable.yml file to the branch to
disable Shippable. But first, we need to keep our docs automation to not
remove the file automatically.

Notice how all shippable builds on `gh-pages` fail: https://app.shippable.com/github/gro-intelligence/api-client/dashboard/history
Step 2 is here: #289


muzigao (approved):
LGTM

PR #262, opened 2020-10-19
CLEWS-32775 fix/workaround for `fetch() called on closed AsyncHTTPClient` errors

jli:
#257 added a __del__ method to GroClient.

After this PR, we've noticed occasional, difficult-to-isolate/reproduce
bugs where GroClient.__del__ is called on objects that are still in use.
GroClient.__del__ includes _async_http_client.close(), which causes
subsequent uses of _async_http_client to raise RuntimeError("fetch()
called on closed AsyncHTTPClient").

FWIW, the spurious calls to __del__ seem occur when:
- GroClient is used in a Celery task
- the task also connects to databases (pandas.io.sql.read_sql_query, psychopg2)

PR #293, opened 2020-12-02
similar regions: improve for multiple runs on same instance

nemozen:
- use correct ball tree in memory  for multiple runs: ball tree is only saved by level so the wrong tree can be used if we keep the same level and change different search_regions. This is a quick fix. Added TODO to make it more efficient

- use a fresh async batch for each build(): clients are very lightweight so this doesn't cost anything but avoids confusing async log messages due to reusing connection pool and queue from previous runs

### DIFF
--- a/api/client/samples/similar_regions/similar_region.py
+++ b/api/client/samples/similar_regions/similar_region.py
@@ -444,8 +444,9 @@ class SimilarRegion(object):
         self._logger.warning("Getting data series for {} regions in {} batch(es) for property {}".
                           format(n_reg, n_queries, prop_name))
         load_bar = tqdm(total=n_reg)
-        # Sending out all queries
-        self.client.batch_async_get_data_points(queries, map_result=map_response)
+        # Sending out all queries -- use fresh client instance to avoid mixing async connection pools between runs
+        data_client = GroClient(API_HOST, ACCESS_TOKEN)
+        data_client.batch_async_get_data_points(queries, map_result=map_response)
         load_bar.close()
         data_counters[data_counters == 0] = np.nan # avoid division by zero warning, where no data
         valid_data = valid_data/data_counters
--- a/api/client/samples/similar_regions/similar_region.py
+++ b/api/client/samples/similar_regions/similar_region.py
@@ -150,11 +150,11 @@ class SimilarRegion(object):
         self._logger.info("Constructing BallTrees...")
         self.metric_object = DistanceMetric.get_metric('pyfunc', **{'func':dist_function})
         self.ball = BallTree(self.data, metric=self.metric_object, leaf_size=2)
+        # TODO: don't reset the following dicts, keep ball trees and regions keyed by (search_region, level) instead of
+        # only level for more reuse
         self.balls_on_level = {}
         self.regions_on_level = {}
-        # separate ball trees to process requests on givel level
-        # shouldn't generally make separate data copies since region_id's are organized with countries < provinces < districts
-        # so data for each level should be continuous arrays
+        self.already_built = set()
         for level in region_levels:
             self.regions_on_level[level] = [r for r in self.available_regions if self.region_info[r]['level']==level]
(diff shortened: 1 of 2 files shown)


jli:
> use a fresh async batch for each build(): clients are very lightweight so this doesn't cost anything but avoids confusing async log messages due to reusing connection pool and queue from previous runs

Just curious, what do the confusing errors look like? Could we also work around this by changing the logging to be less verbose?

Agreed it should be cheap, though just a head's up that we've run into some strange issues w/ creating many client objects in the past: #257 #262

jli (approved):

api/client/samples/similar_regions/similar_region.py
```diff
@@ -444,8 +444,9 @@ def map_response(idx, _, response, *args):
         self._logger.warning("Getting data series for {} regions in {} batch(es) for property {}".
                           format(n_reg, n_queries, prop_name))
         load_bar = tqdm(total=n_reg)
-        # Sending out all queries
-        self.client.batch_async_get_data_points(queries, map_result=map_response)
+        # Sending out all queries -- use fresh client instance to avoid mixing async connection pools between runs
+        data_client = GroClient(API_HOST, ACCESS_TOKEN)
```
ash-gro, line 448: Just to make sure I understand - this data_client is local and separate from self.client. This is only used for this particular data download (i.e. this particular property and region list). I don't think I saw "confusing messages" you are referring to and not sure how having a separate client would help. But if it is I guess this is fine

api/client/samples/similar_regions/similar_region.py
```diff
@@ -150,11 +150,11 @@ def dist_function(x,y):
...
         self.metric_object = DistanceMetric.get_metric('pyfunc', **{'func':dist_function})
         self.ball = BallTree(self.data, metric=self.metric_object, leaf_size=2)
+        # TODO: don't reset the following dicts, keep ball trees and regions keyed by (search_region, level) instead of
+        # only level for more reuse
         self.balls_on_level = {}
         self.regions_on_level = {}
-        # separate ball trees to process requests on givel level
-        # shouldn't generally make separate data copies since region_id's are organized with countries < provinces < districts
-        # so data for each level should be continuous arrays
+        self.already_built = set()
```
ash-gro, line 157: We rebuild if search region OR requested_level are not present in already_bult. If it is search_region which changed, we should clear everything (all trees on all levels become invalid). If it is only requested_level, we can afford to keep other trees around. With this change we clear all trees regardless. This might be less efficient but should be safe.

ash-gro (approved):

nemozen:
> > use a fresh async batch for each build(): clients are very lightweight so this doesn't cost anything but avoids confusing async log messages due to reusing connection pool and queue from previous runs
> 
> Just curious, what do the confusing errors look like? Could we also work around this by changing the logging to be less verbose?
> 
> Agreed it should be cheap, though just a head's up that we've run into some strange issues w/ creating many client objects in the past: #257 #262



> > use a fresh async batch for each build(): clients are very lightweight so this doesn't cost anything but avoids confusing async log messages due to reusing connection pool and queue from previous runs
> 
> Just curious, what do the confusing errors look like? Could we also work around this by changing the logging to be less verbose?
> 

When running in a ipynb with nest_asyncio (as in example.ipynb) andyou stop a cell before it finishes, and then start another cell, you occasionally get messages from the load_bar from the previous cell mixed with the current one.

Also, when connections are reset from the server, the client instance continues to get connection reset messages on subsequent async_get_data calls. 

I think there are some improvement to be made on the client's io queue handling


> Agreed it should be cheap, though just a head's up that we've run into some strange issues w/ creating many client objects in the past: #257 #262

api/client/samples/similar_regions/similar_region.py
nemozen: yeah, the point is  I think we can avoid rebuilding if search region is changed but for now that is the case. see TODO comment above.

PR #295, opened 2020-12-04
Drop points with missing data when loading file to prevent nan propagation

ash-gro:
A rather trivial change. If user data file has missing values they will be interpreted as nan's and propagated by numpy operations (specifically sum(axis=0) ) to the output. One nan is enough to ruin entire output. The easiest solution is just to drop such points immediately upon load which is done here.

### DIFF
--- a/api/client/samples/similar_regions/similar_region.py
+++ b/api/client/samples/similar_regions/similar_region.py
@@ -350,7 +350,7 @@ class SimilarRegion(object):
         # will return array of long-term averages for each time interval within a year, even for static
         result = np.zeros(self.t_int_per_year)
         try:
-            raw_data = pd.read_csv(full_path, parse_dates=[0,1], infer_datetime_format=True)
+            raw_data = pd.read_csv(full_path, parse_dates=[0,1], infer_datetime_format=True).dropna()
         except Exception as e:
             self._logger.error("Can not get data from file {}: {}".format(full_path, str(e)))
             return None, [] # empty list signals no data
--- a/api/client/samples/similar_regions/similar_region.py
+++ b/api/client/samples/similar_regions/similar_region.py
@@ -351,7 +351,7 @@ class SimilarRegion(object):
         result = np.zeros(self.t_int_per_year)
         try:
             raw_data = pd.read_csv(full_path, parse_dates=[0,1], infer_datetime_format=True)
-            num_rows_with_na = raw_data.isna().sum().max()
+            num_rows_with_na = (raw_data.isna().sum(axis=1)>0).sum() #raw_data.isna().sum().max()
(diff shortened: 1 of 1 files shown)


api/client/samples/similar_regions/similar_region.py
```diff
@@ -350,7 +350,7 @@ def _load_property_from_file(self, prop_name, path_prefix):
         # will return array of long-term averages for each time interval within a year, even for static
         result = np.zeros(self.t_int_per_year)
         try:
-            raw_data = pd.read_csv(full_path, parse_dates=[0,1], infer_datetime_format=True)
+            raw_data = pd.read_csv(full_path, parse_dates=[0,1], infer_datetime_format=True).dropna()
```
jli, line 353: I think it's nice to be explicit w/ data cleaning operations that are happening automatically - for instance, in this case, I think it would be nice to print a message that N rows are being dropped due to containing nulls.

Something like this:

```python
            raw_data = pd.read_csv(full_path, parse_dates=[0,1], infer_datetime_format=True)
            num_rows_with_na = raw_data.isna().sum().max()
            if num_rows_with_na:
                self._logger.warn(f"Note: dropping {num_rows_with_na} rows that contain null values.")
                raw_data.dropna(inplace=True)
```

WDYT?

nemozen (commented):
Change looks good, just one minor comment..

Not related to the function being changed here itself but as a future improvement, we can merge the logic for the seed region and the search region, and then  it would be nice to have the side-loading from file work in general for any region.

api/client/samples/similar_regions/similar_region.py
ash-gro: Rather significant amount of missing data can be tolerated (as long as there is at least one valid data point per interval), but sure, it does not hurt, I guess... Done in the next commit.

jli (approved):

api/client/samples/similar_regions/similar_region.py
```diff
@@ -350,7 +350,7 @@ def _load_property_from_file(self, prop_name, path_prefix):
...
         except Exception as e:
             self._logger.error("Can not get data from file {}: {}".format(full_path, str(e)))
             return None, [] # empty list signals no data
```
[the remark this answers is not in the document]
ash-gro: As always with error handling this depends on preference/style, I think. We have two different ways of loading data - from Gro/cache and a file. Gro/cache variant of loading already returns empty list of valid regions if it does not get data (and can't just throw unhandled Exception since we load incrementally and might get data for other regions), so I just force the same processing by caller by returning empty valid region list here as well.

nemozen (approved):

Issue #290, opened 2020-11-30
import issue with brazil_soybeans.ipynb


sahuguet:
The import needs to be changed to
`from groclient.crop_model import CropModel`.

https://github.com/gro-intelligence/api-client/blob/development/api/client/samples/crop_models/brazil_soybeans.ipynb

nemozen:
FYI I reran the notebook as is with the  latest version of the client and the backward compatibility shim https://github.com/gro-intelligence/api-client/blob/development/api/client/crop_model/__init__.py  is behaving as expected, i..e logs a warning but the code works.

jli:
The import was updated in #296.

PR #296, opened 2020-12-07
Add script to test subset of Jupyter notebooks with nbval plugin (CLEWS-33407)

jli:
This adds `bin/test_notebooks.sh`, a script that tests a subset of the sample notebooks using [nbval](https://github.com/computationalmodelling/nbval). "Lax" testing just checks that the notebook code executes without errors (catches broken imports, unspecified requirements, basic code/syntax issues, etc.). "Strict" testing also checks that the output for each notebook cell matches what was recorded in the notebook originally.

Currently, 2 notebooks pass strict testing and 2 notebooks pass lax testing.

This PR fixes simple problems with several notebook, including wrong groclient import (#290) and similar issues with several other notebooks.

There are 7 other notebooks in `api/client/samples` that currently can't be tested:
- code error (we should fix these)
  - analogous_years/get_started_with_analogous_years.ipynb: throws exception. numpy shape issue
  - crop_models/Sao_Paulo_brazil_sugar_content_modeling.ipynb: a cell references a value only defined in a later cell.
- requirements challenges (e.g., need to install a system library)
  - crop_models/brazil_soybeans.ipynb: requirements: requires gdal system library?
  - prevented_plant/prevented_plant_models.ipynb: requirements: Shapely requires geos_c library?
  - similar_regions/example.ipynb: requirements: geopandas requires gdal stuff
- runtime speed: these seem to query a lot of data and take a long time to execute
  - crop_models/ethiopia_cereals.ipynb
  - crop_models/kenya_cereals.ipynb


api/client/samples/analysis_kits/stocks_to_use_price_model/requirements.txt
```diff
@@ -1,4 +1,4 @@
-pandas~=0.24.2
-matplotlib~=3.1.1
-scikit-learn~=0.21.2
-jupyter
\ No newline at end of file
+pandas
```
muzigao, line 1: @jli does the version matter here?

muzigao (approved):

api/client/samples/analysis_kits/stocks_to_use_price_model/requirements.txt
jli: Summary of offline discussion: I originally removed the version restriction because I had trouble installing these dependencies locally (I think that version of Pandas required building numpy locally, which failed?). I figured it was fine to remove, since almost none of the other sample notebook requirements.txt files have version restrictions.

It may be good to at least specify a major version to avoid breakages from new library versions, though there it also a benefit to not have version restrictions (for tested notebooks) since we'll be able to keep testing that our code works with the latest library versions.

PR #297, opened 2020-12-08
fix: each GroClient has its own AsyncHTTPClient (CLEWS-32775)

jli:
This should finally, properly fix the `fetch() called on closed
AsyncHTTPClient` errors that we've been seeing since #257. The root
problem is that AsyncHTTPClient has an internal instance cache, so
creating a new GroClient when a previous one exists would reuse the
internal AsyncHTTPClient. If the previous GroClient falls out of scope,
the (shared) AsyncHTTPClient is closed (via GroClient.__del__),
resulting in errors when trying to use the new GroClient.

- See the Tornado docs for `AsyncHTTPClient`, `force_instance`, and the
  `close` method: https://www.tornadoweb.org/en/stable/httpclient.html#tornado.httpclient.AsyncHTTPClient
- We explicitly started closing the AsyncHTTPClient in this PR (due to
  file descriptor leaking): https://github.com/gro-intelligence/api-client/pull/257
- This previous workaround fixed some instances of the error on Celery: https://github.com/gro-intelligence/api-client/pull/262
  It's still mysterious why that PR fixed things, and why more things on
  Celery weren't failing... Perhaps Celery does other things with
  IOLoops that result in this bug not being triggered sometimes.

### DIFF
--- a/groclient/batch_client_test.py
+++ b/groclient/batch_client_test.py
@@ -122,6 +122,18 @@ class BatchTests(TestCase):
 
         self.assertEqual(summation, 97680)
 
+    # Test that multiple GroClients each have their own AsyncHTTPClient. Note:
+    # this tests the fix for the `fetch called on closed AsyncHTTPClient`
+    # error. We can't test for that directly since the `fetch` call is mocked,
+    # so instead we just ensure that all GroClients have their own
+    # AsyncHTTPClient.
+    def test_batch_async_get_data_points_multiple_clients(self):
+        client = GroClient(MOCK_HOST, MOCK_TOKEN)
+        ahc_id1 = id(client._async_http_client)
+        client = GroClient(MOCK_HOST, MOCK_TOKEN)
+        ahc_id2 = id(client._async_http_client)
+        self.assertNotEqual(ahc_id1, ahc_id2)
+
     def test_batch_async_get_data_points_bad_request_error(self):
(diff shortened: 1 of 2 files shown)


tolothy (approved):
Logic is solid. And my test script which failed due to `fetch() called on closed AsyncHTTPClient` now works with this fix
```
from gro_utils import api_client_helpers

api_host = api_client_helpers.get_default_api_client_host()
api_token = api_client_helpers.get_default_access_token()

input_list = [
    {'metric_id': 860032, 'item_id': 274, 'region_id': 1215, 'frequency_id': 9, 'source_id': 2},
    {'metric_id': 860032, 'item_id': 270, 'region_id': 1215, 'frequency_id': 9, 'source_id': 2}
]
output_list = []
def map_response(inputIndex, inputObject, response, output_list):
    output_list += response
    return output_list

client = GroClient(api_host, api_token)
batch_output = client.batch_async_get_data_points(input_list, output_list=output_list, map_result=map_response)
print(batch_output[:2])

client = GroClient(api_host, api_token)
batch_output = client.batch_async_get_data_points(input_list, output_list=output_list, map_result=map_response)
print(batch_output[:2])
```

eliaschneider (approved):
LGTM, 

I have only one quick question: is this changing reducing performance? querying multiple regions with batch_async_get_data_points will be slower?

jli:
@eliaschneider There shouldn't be any performance implications.

PR #299, opened 2020-12-30
fix at_time notebook links

jli:
Note: the "source lag" link was replaced with a link to the FAQ doc in our API developer docs, but there wasn't a corresponding question in that doc for the 2nd link: "how do I see previous values fo a time series point to see how the value changed over time".

If this functionality is still possible, we should probably add it to the current FAQ doc?

### DIFF
--- a/api/client/samples/at-time-query-examples.ipynb
+++ b/api/client/samples/at-time-query-examples.ipynb
@@ -47,7 +47,8 @@
     "--------\n",
     "- https://developers.gro-intelligence.com/api.html#groclient.GroClient.get_data_points\n",
     "- https://developers.gro-intelligence.com/api.html#groclient.GroClient.lookup\n",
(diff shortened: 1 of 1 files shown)


muzigao:
@jli  the second point is the "show_revisions" flag in from https://developers.gro-intelligence.com/api.html#groclient.GroClient.get_data_points

@JGrinst   do you remember if we had "how do I see previous values fo a time series point to see how the value changed over time" in FAQ before? was that removed on purpose?

JGrinst:
<p>We deleted the github wiki, so that we wouldn't have documentation in two different places. But I believe that specific FAQ is now covered by https://developers.gro-intelligence.com/Recipes/querying-data.html#show-revisions</p>

jli:
Thanks Justin! I replaced the 2nd wiki link with that link for `show_revisions`.

Muzi, please take another look.

nemozen (approved):

muzigao (approved):

PR #300, opened 2021-01-07
feat: make GroClient() args optional

jli:
Changes `GroClient.__init__` so that both API host and access token are optional arguments. The API host defaults to the production API server and the access token defaults to `$GROAPI_TOKEN`. This saves clients (and us) from needing to spell out 'api.gro-intelligence.com' over and over.

Also added docstring for the constructor and added to developer docs: https://developers.gro-intelligence.com/jli-convenient-constructor/api.html#creating-a-client

### DIFF
--- a/groclient/client.py
+++ b/groclient/client.py
@@ -82,6 +82,11 @@ class GroClient(object):
 
             >>> client = GroClient(access_token="your_token_here")
         """
+        # Initialize early since they're referenced in the destructor and
+        # access_token checking may cause constructor to exit early.
+        self._async_http_client = None
+        self._ioloop = None
+
         if access_token is None:
             access_token = os.environ.get("GROAPI_TOKEN")
             if access_token is None:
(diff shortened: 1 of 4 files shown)


groclient/client.py
```diff
@@ -57,7 +58,36 @@ def __init__(self, response, retry_count, url, params):
...
+                :code:`$GROAPI_TOKEN` environment variable are set.
+
+        Examples
+        --------
+            >>> client = GroClient()  # token stored in $GROAPI_TOKEN
+
+            >>> client = GroClient(access_token="your_token_here")
+        """
+        if access_token is None:
+            env_token = os.environ.get("GROAPI_TOKEN")
```
nemozen: we don't really need a separate variable here,  just use access_token rather than setting it on line 90

nemozen (commented):
lgtm just one minor comment

groclient/client.py
jli: done

PR #301, opened 2021-01-19
CLEWS-35217 include available dates in get data call

cn1036:
**Prerequisite:** 
https://bitbucket.org/grointelligence/gro/pull-requests/19010/clews-35217-include-available-dates-in-api

**tested locally**
Untitled


groclient/lib.py
```diff
@@ -488,13 +489,15 @@ def list_of_series_to_single_series(series_list, add_belongs_to=False, include_h
                 'end_date': point[1],
                 'value': point[2],
                 'unit_id': point[4] if len(point) > 4 else series['series'].get('unitId', None),
-                'metadata': point[5] if len(point) > 5 else {},
+                'metadata': point[5] if len(point) > 5 and point[5]!=None else {},
```
jli: nit: `is not None` is preferred (2nd point here: https://www.python.org/dev/peps/pep-0008/#programming-recommendations)

For older clients, will they have a `metadata` value of `None` instead of `{}` after the API server change?

groclient/client.py
```diff
@@ -902,7 +902,7 @@ def get_top(self, entity_type, num_results=5, **selection):
             self.access_token, self.api_host, entity_type, num_results, **selection
         )
 
-    def get_df(self, show_revisions=False, index_by_series=False, include_names=False):
+    def get_df(self, show_revisions=False, show_available_date=False, index_by_series=False, include_names=False):
```
jli, line 905: I know it was already out of date... but would you mind updating the doc string? :)

jli (commented):
Could you also update `docs/data-point-definition.md`?

It includes documentation for all the fields returned by get_data_points: https://developers.gro-intelligence.com/data-point-definition.html

jli:
Compatibility question: does this code behave properly with the current API server (without your change)? (Even if we need to roll back the API server change, clients with the latest code should keep working.)

groclient/lib.py
cn1036: updated to use 'is not None'

For the second question, clients on older version will still get {}. Users will not be able to turn on the showAvailableDate option if they don't upgrade to the newer version, so availableDate is always null. Since we are removing trailing Nulls in each data point(dataUtils - flattenTimeSeriesPoint()), the API response becomes like this:
1) if metadata != Null
[{series: {...}, data: [[start, end, value, reportingDate, unit, metadata], ...]}]
2) if metadata = Null, it will be deleted together with availableDate
[{series: {...}, data: [[start, end, value, reportingDate, unit], ...]}]

Thus for current version, the statement _len(point) > 5_ is equivalent to _metadata is not None_

cn1036:
> Compatibility question: does this code behave properly with the current API server (without your change)? (Even if we need to roll back the API server change, clients with the latest code should keep working.)

No. Once we expose the show_available_date option to API clients, the v2/data endpoint must accept it as a param, otherwise user will get following error msg
image

groclient/client.py
cn1036: good catch! updated

cn1036:
> Could you also update `docs/data-point-definition.md`?
> 
> It includes documentation for all the fields returned by get_data_points: https://developers.gro-intelligence.com/data-point-definition.html

done. Added definition for 'metadata' and 'available_date', please take a look

docs/data-point-definition.md
```diff
@@ -24,7 +26,9 @@ Below are some explanations of what each of those fields represent:
 * `end_date`: end of the period this point represents
 * `value`: the value, typically a number. In some cases, the value may be non-numeric. E.g., when the metric is Crop Calendar, a value of "planting," represents the fact that the planting period is from `start_date` to `end_date`.
 * `unit_id`: this is a Gro unit id you can look up the name/abbreviation/etc. of using the `client.lookup('units', unit_id)` function. There's also a helper function of which you can see an example in the [quickstart](https://github.com/gro-intelligence/api-client/blob/9c2c17642980b5415b8a8167a28276b77e34915c/api/client/samples/quick_start.py#L30) for getting just the abbreviation from the unit id, `client.lookup_unit_abbreviation(point['unit_id'])`, which is the common case you probably want
-* `reporting_date`: date the source reported this value (only included when source provides reporting date)
+* `reporting_date`: date the source reported this value (defined only when source provides reporting date)
+* `metadata`: additional information (i.e. {"conf_interval": 0.7})
+* `available_date`: date this value became stable in Gro system
```
muzigao: @JGrinst  any thoughts on this language?

JGrinst: This looks pretty good, and is consistent with what we state elsewhere.

cn1036:
this PR has been split into small PRs. closing

PR #303, opened 2021-01-26
CLEWS-35446 example notebook showing available_date on the stage environment

jli:
still in progress!

You can see the notebook on GitHub: https://github.com/gro-intelligence/api-client/blob/CLEWS-35446-stage-notebook/api/client/samples/stage_environment_example.ipynb

### DIFF
--- a/api/client/samples/stage_environment_available_date.ipynb
+++ b/api/client/samples/stage_environment_available_date.ipynb
@@ -1,0 +1,384 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Requesting available_date from Gro's \"stage\" API\n",
+    "\n",
+    "This notebook demonstrates:\n",
+    "1. how to connect to Gro's \"stage\" environment\n",
+    "2. how to use the `show_available_date` option."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Connecting to the Gro \"stage\" API\n",
+    "\n",
+    "Connecting to the \"stage\" API environment instead of prod is configured when constructing the `GroClient` object. `GroClient` takes 2 arguments: the API domain name and your account's API access token.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


docs/data-point-definition.md
```diff
@@ -8,6 +8,8 @@ Gro defines a "data point" as a discrete result produced by our API. When using
   'value': 0.131714797651,
   'unit_id': 851,
   'reporting_date': None,
+  'metadata': {},
+  'available_date': None,
```
doriapenciucgro: I see, we are pushing both the documentation updates and stage examples in the same PR

docs/data-point-definition.md
```diff
@@ -24,7 +26,9 @@ Below are some explanations of what each of those fields represent:
 * `end_date`: end of the period this point represents
 * `value`: the value, typically a number. In some cases, the value may be non-numeric. E.g., when the metric is Crop Calendar, a value of "planting," represents the fact that the planting period is from `start_date` to `end_date`.
 * `unit_id`: this is a Gro unit id you can look up the name/abbreviation/etc. of using the `client.lookup('units', unit_id)` function. There's also a helper function of which you can see an example in the [quickstart](https://github.com/gro-intelligence/api-client/blob/9c2c17642980b5415b8a8167a28276b77e34915c/api/client/samples/quick_start.py#L30) for getting just the abbreviation from the unit id, `client.lookup_unit_abbreviation(point['unit_id'])`, which is the common case you probably want
-* `reporting_date`: date the source reported this value (only included when source provides reporting date)
+* `reporting_date`: date the source reported this value (defined only when source provides reporting date)
```
doriapenciucgro: Should it say "date when"? Same for available_date?
I wonder if this would be more clear if we said: 
"date when the source reported this value (returns None if that information is not available)"

doriapenciucgro (approved):
Added a couple of minor questions, otherwise looks good to me

docs/data-point-definition.md
jli: I changed this branch to drop the documentation updates now.

docs/data-point-definition.md
jli: Thanks, that sounds better to me! I'll update #301 with these changes.

PR #305, opened 2021-01-28
CLEWS-35217 only include available dates if requested by user

cn1036:
Discussion: https://gro-intelligence.slack.com/archives/C01JFH4DK62/p1611860350044100

tested locally
image

### DIFF
--- a/groclient/client.py
+++ b/groclient/client.py
@@ -982,7 +982,8 @@ class GroClient(object):
         tmp.end_date = pandas.to_datetime(tmp.end_date)
         tmp.start_date = pandas.to_datetime(tmp.start_date)
         tmp.reporting_date = pandas.to_datetime(tmp.reporting_date)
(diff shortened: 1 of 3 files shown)


groclient/client.py
```diff
@@ -282,8 +282,9 @@ def get_data_points_generator(self, **selection):
         try:
             list_of_series_points = yield self.async_get_data(url, headers, params)
             include_historical = selection.get("include_historical", True)
+            include_available_date = selection.get('show_available_date', False)
             points = lib.list_of_series_to_single_series(
-                list_of_series_points, False, include_historical
+                list_of_series_points, False, include_historical. include_available_date
```
doriapenciucgro: @nicheng22 should the period here be a comma instead?

cn1036: @doriapenciucgro good catch! fixed

jli (approved):

PR #308, opened 2021-01-30
CLEWS-35217: remove available_date from documentation

muzigao:
Hide new available_date param from get_data_points() until it's officially released


doriapenciucgro (approved):
Looks good

PR #310, opened 2021-02-05
Update doc

cn1036:
Add comments on available_date

### DIFF
--- a/docs/data-point-definition.md
+++ b/docs/data-point-definition.md
@@ -9,7 +9,7 @@ Gro defines a "data point" as a discrete result produced by our API. When using 
   'unit_id': 851,
   'reporting_date': None,
   'metadata': {},
-  'available_date': None,
+  'available_date': '2013-03-01T00:00:00.000Z',
   'metric_id': 15531082,
   'item_id': 7382,
   'region_id': 138295,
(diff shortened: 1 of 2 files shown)


doriapenciucgro (approved):

groclient/client.py
```diff
@@ -1095,6 +1100,8 @@ def get_data_points(self, **selections):
         show_revisions : boolean, optional
             False by default, meaning only the latest value for each period. If true, will return
             all values for a given period, differentiated by the `reporting_date` field.
+        show_available_date : boolean, optional
```
muzigao, line 1103: @cn1036  could we also update output examples? I took out  those reference in previous PR https://github.com/gro-intelligence/api-client/pull/308/files

docs/data-point-definition.md
```diff
@@ -8,6 +8,8 @@ Gro defines a "data point" as a discrete result produced by our API. When using
   'value': 0.131714797651,
   'unit_id': 851,
   'reporting_date': None,
+  'metadata': {},
+  'available_date': None,
```
jli: `None` shouldn't be a valid value of `available_date` anymore, right? It should either not appear in the result at all (if it wasn't requested), or it should have a timestamp string as the value?

jli:
Is #301 still relevant, or can we close that PR?

docs/data-point-definition.md
cn1036: good catch! updated with actual timestamp

cn1036:
> Is #301 still relevant, or can we close that PR?

closed

jli (approved):

muzigao (changes requested):

groclient/client.py
muzigao: ha never mind, since we removed available_date completely when it's not specified

muzigao (approved):

PR #317, opened 2021-04-22
Add circleci config

kmetallic04:
Add relevant configuration files for circle ci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
     # ref: https://circleci.com/docs/2.0/executor-types/?section=executors-and-images#using-docker
     docker:
       # Use the base circle CI image
-      - image: "cimg/base:stable"
+      - image: "cimg/python:3.7"
     steps:
       # TODO set up source and pip caching
       # checkout the git repo
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
 
       - run:
           name: "Run tests and code coverage"
-          command: |
+          command: >-
             mkdir -p circleci/testresults circleci/codecoverage &&
             poetry run pytest
             --junitxml=circleci/testresults/junit.xml
@@ -41,7 +41,7 @@ jobs:
       
       - run:
           name: "Test sample notebook"
-          command: |
+          command: >-
             pip install -r api/client/samples/analogous_years/requirements.txt &&
             poetry run pytest api/client/samples/analogous_years/
 
@@ -50,7 +50,7 @@ jobs:
       # https://circleci.com/docs/2.0/configuration-reference/#run
       - run:
           name: "Ensure latest docs build without warnings or errors"
-          command: |
+          command: >-
             poetry install -E docs &&
             poetry run sphinx-build -W --keep-going docs docs/_build/html
 
@@ -63,7 +63,7 @@ jobs:
       # See note in CONTRIBUTING.md
       - run:
           name: "Build package"
-          command: |
+          command: >-
             pip install poetry-dynamic-versioning &&
             poetry build
 
@@ -73,7 +73,7 @@ jobs:
       # merged, which is why we check $IS_PULL_REQUEST.
       - run:
           name: "Check for broken links in the docs"
-          command: |
+          command: >-
             if [ "$CIRCLE_BRANCH" == "development" ] && [ ! -z $CIRCLE_PULL_REQUEST ]; then
               poetry config repositories.testpypi https://test.pypi.org/legacy/ &&
               poetry publish -u __token__ -p $TESTPYPI_TOKEN -r testpypi
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,13 +1,59 @@
-# Use the latest 2.1 version of CircleCI pipeline process engine. See: https://circleci.com/docs/2.0/configuration-reference
+---
+
 version: 2.1
-# Use a package of configuration called an orb.
-orbs:
-  # Declare a dependency on the welcome-orb
-  welcome: circleci/welcome-orb@0.4.1
-# Orchestrate or schedule a set of jobs
+
+jobs:
+  ci:
+    # ref: https://circleci.com/docs/2.0/executor-types/?section=executors-and-images#using-docker
+    docker:
+      # Use the base circle CI image
+      - image: "cimg/base:stable"
+        auth:
+          username: DOCKERHUB_LOGIN
+          password: DOCKERHUB_PASSWORD
+    steps:
+      # TODO set up source and pip caching
+      # checkout the git repo
+      - checkout
+
+      - run:
+          name: "Run tests and code coverage"
+          command: |
+            pip install -U pip wheel
+            pip install poetry
+            poetry install
+            poetry run python groclient/lib.py -v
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1,13 +1,59 @@
...
+          command: |
+            pip install -U pip wheel
+            pip install poetry
+            poetry install
+            poetry run python groclient/lib.py -v
+            mkdir -p shippable/testresults shippable/codecoverage &&
+            poetry run pytest
+            --junitxml=shippable/testresults/nosetests.xml
+            --cov=groclient
+            --cov-report=xml:shippable/codecoverage/coverage.xml
```
jli: Let's remove --junitxml and --cov-report, since those seem tailored to shippable?

.circleci/config.yml
```diff
@@ -1,13 +1,59 @@
...
+          name: "Run post build steps"
+          when: on_success
+          command: |
+            poetry install -E docs &&
+            poetry run sphinx-build -W --keep-going docs docs/_build/html
+            poetry run sphinx-build -b linkcheck docs docs/_build/linkcheck
+            git config --global user.email "api-documentation@gro-intelligence.com" &&
+            git config --global user.name "Gro Intelligence" &&
+            git remote set-url origin git@github.com:<< pipeline.project.git_url >> &&
+            poetry run bin/sphinx_push_ghpages.sh
```
jli: Should we avoid doing this for now, since it will conflict with shippable?

.circleci/config.yml
```diff
@@ -1,13 +1,59 @@
...
+            pip install poetry-dynamic-versioning &&
+            poetry build
+            if [ "$BRANCH" == "development" -a "$IS_PULL_REQUEST" == "false" ]; then
+              poetry config repositories.testpypi https://test.pypi.org/legacy/ &&
+              poetry publish -u __token__ -p $TESTPYPI_TOKEN -r testpypi
+            fi
+            if [ "$IS_RELEASE" == "true" ]; then
+              poetry publish -u __token__ -p $PYPI_TOKEN
```
jli: I think we should probably avoid doing this (for now) since it'll conflict with Shippable?

Publishing to testpypi seems fine, though.

jli: Also, how are the TESTPYPI_TOKEN and PYPI_TOKEN env vars getting set in CircleCI?

.circleci/config.yml
```diff
@@ -1,13 +1,59 @@
...
+            mkdir -p shippable/testresults shippable/codecoverage &&
+            poetry run pytest
+            --junitxml=shippable/testresults/nosetests.xml
+            --cov=groclient
+            --cov-report=xml:shippable/codecoverage/coverage.xml
+            pip install -r api/client/samples/analogous_years/requirements.txt &&
+            poetry run pytest api/client/samples/analogous_years/
+
+      - run:
+          name: "Run post build steps"
```
jli: The shippable.yml had a lot of comments here, which I think is useful to retain.

.circleci/config.yml
kmetallic04: Hmm.... I think a rename would be appropriate, since we would still require to upload and persist test results + coverage, as described here: https://circleci.com/docs/2.0/collect-test-data/#pytest and here: https://circleci.com/docs/2.0/code-coverage/#python. Updated and stored the results as artifacts. WDYT?

.circleci/config.yml
kmetallic04: Right. Fixed.

.circleci/config.yml
kmetallic04: This was just a debugging draft of the PR, should have done a better job of putting that in the description. I haven't yet set up the environment variables, but I'm planning to use the following measures for open source projects' variables: https://circleci.com/docs/2.0/oss/

kmetallic04: Removed the line to publish to pypi, leaving testpypi.

.circleci/config.yml
kmetallic04: Done.

kmetallic04:
See a passing Circle CI pipeline here: https://app.circleci.com/pipelines/github/kmetallic04/api-client/13/workflows/c4fe7f4c-f71f-4545-aeb9-62b256168074/jobs/17/steps

.circleci/config.yml
```diff
@@ -1,13 +1,89 @@
...
+          name: "Check for broken links in the docs"
+          command: |
+            poetry run sphinx-build -b linkcheck docs docs/_build/linkcheck
+      
+      # Note: We need to install versioning library outside of the poetry venv.
+      # See note in CONTRIBUTING.md
+      - run:
+          name: "Build package"
+          command: >-
+            poetry run pip install poetry-dynamic-versioning &&
```
jli: I think this needs to be just `pip install`, not `poetry run pip install`.

IIRC, the dynamic versioning library needs to be available in the external environment for poetry to make use of it.

You can see here that the package has version `0.0.0`: https://app.circleci.com/pipelines/github/kmetallic04/api-client/13/workflows/c4fe7f4c-f71f-4545-aeb9-62b256168074/jobs/17/parallel-runs/0/steps/0-110

The version should reference the latest tag. For most builds, it includes a "dev" suffix indicating how many commits on top of the most recent tag there are (e.g. "1.122.0.dev123", see https://test.pypi.org/project/groclient/#history). For tagged release  builds, there are 0 extra commits, so the version is just the tag value exactly.

.circleci/config.yml
jli: Perhaps include the command in a TODO comment?

jli:
Looks good!

1. Main issue is the poetry dynamic versioning library.
2. How do we configure Github/CircleCI so that it runs on every pull request, like Shippable does now?

Other things to test: publishing Sphinx docs and publishing to PyPI. I think after we address the 2 issues above, we could disable the Sphinx and PyPI parts of Shippable and enable them on CircleCI to test that functionality.

.circleci/config.yml
kmetallic04: Ah, thanks for the catch! Fixed.

kmetallic04:
@jli , I think the configuration to let CircleCI run on every PR will require some admin rights, to enable GitHub checks, GitHub status and to set up the environment variables. We also need to build pull requests from forked repos as described here: https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories. I'll be following up on this.

.circleci/config.yml
```diff
@@ -1,13 +1,90 @@
...
+      # Publish new development updates to TestPyPI whenever changes are merged
+      # to development.
+      # We only want to publish a new package when PRs are actually accepted and
+      # merged, which is why we check $CIRCLE_PULL_REQUEST.
+      - run:
+          name: "Push to testpypi"
+          command: >-
+            if [ "$CIRCLE_BRANCH" == "development" ] && [ ! -z $CIRCLE_PULL_REQUEST ]; then
+              poetry config repositories.testpypi https://test.pypi.org/legacy/ &&
+              poetry publish -u __token__ -p $TESTPYPI_TOKEN -r testpypi
```
jli: we should probably remove this for now, as the token isn't configured yet?

jli (changes requested):

jli (approved):

PR #321, opened 2021-05-12
CLEWS-38092:  support extra params in async_get_df()

muzigao:
Extend current `async_get_df()` function to support extra parameters like show_revisions, include_names etc


eliaschneider (approved):
LGTM

muzigao:
added `async_mode` to `get_df()` parameters so we can consolidate `get_df()` and `async_get_df()` now

groclient/client.py
```diff
@@ -958,60 +965,76 @@ def get_df(self, show_revisions=False, show_available_date=False, index_by_serie
...
             for entity_type_id in DATA_SERIES_UNIQUE_TYPES_ID + ['unit_id']:
                 name_col = entity_type_id.replace('_id', '_name')
                 name_cols.append(name_col)
-                self._data_frame[name_col] = self._data_frame[entity_type_id].apply(partial(get_name, entity_type_id))
+                entity_dict = self.lookup(ENTITY_KEY_TO_TYPE[entity_type_id], self._data_frame[entity_type_id].unique())
+                self._data_frame[name_col] = self._data_frame[entity_type_id].apply(lambda entity_id: entity_dict.get(str(entity_id))['name'])
 
             if compress_format:
-                table_df = self._data_frame.pivot_table(index='end_date', values='value',
+                # BUG: index by end_date will compute average for all revisions
```
tolothy: should we change it by taking the latest value by `reporting_date` when pivoting the table? or should we leave it as it is but mention it in the docstring of `compress_format`, e.g. something like **dont set compress_format if show_revisions=True**

groclient/client.py
```diff
@@ -958,60 +965,76 @@ def get_df(self, show_revisions=False, show_available_date=False, index_by_serie
...
             compress_format: boolean, optional
-               If set, each series will be compressed to a single column in the dataframe, with the end_date column 
-               set as the dataframe inde. All the entity names for each series will be 
+               If set, each series will be compressed to a single column in the dataframe, with the end_date column
+               set as the dataframe inde. All the entity names for each series will be
                placed in column headers.
-
+            async_mode: boolean, optional
+                If set, it will make :meth:`~get_data_points` requests asynchronously.
+                Note that if you're using Jupyter Notebook, you will need to find a Tornado workaround
```
nemozen: mention the actual  workaround... e.g. "Note that when running in a Jupyter Ipython notebook with async_mode, you will need to use `nest_asyncio` module "

groclient/client.py
nemozen: yes  averaging across reporting date could be confusing. let's be very clear and just but an assertion
`assert not (compress_format and show_revisions), "show_revisions and compress_format are mutually exclusive`

nemozen (changes requested):
lg minor comments below

tolothy (approved):

groclient/client.py
muzigao: updated

nemozen (approved):

PR #322, opened 2021-05-13
Clews 38306 fix text format

raphael-cohen:
Fixed `\` character showing up instead of new line


jeanzhaoyi (approved):
Looks good to me. I like that you also updated the dates in the cell run results.

PR #324, opened 2021-05-18
Do not run ci builds for gh-pages branch

kmetallic04:
Avoid running ci tests for gh-pages branch to save credits & minimize alert fatigue


muzigao (approved):

bin/sphinx_push_ghpages.sh
```diff
@@ -70,7 +70,7 @@ else
 fi
 
 echo -e "\n-> committing change..."
-git commit -m "sphinx_push_ghpages.sh: autocommit $(date '+%Y-%m-%d %H:%M:%S')"
+git commit -m "[skip ci] sphinx_push_ghpages.sh: autocommit $(date '+%Y-%m-%d %H:%M:%S')"
```
muzigao, line 73: @kmetallic04  just making sure, this will not prevent us from seeing doc changes right?
for example:  https://developers.gro-intelligence.com/CLEWS-38706-doc/api.html#groclient.GroClient.get_descendant  (this section was added on CLEWS-38706-doc branch)

kmetallic04: @muzigao Hmm... This will only prevent CircleCI from attempting to run a build from `gh-pages` whenever anything is pushed to the branch. The docs branch does not have the 'usual' files on `development`, including `.circleci/config.yml`, which means any build will fail right away from the trigger due to missing config files; see https://app.circleci.com/pipelines/github/gro-intelligence/api-client?branch=gh-pages. This happens on every commit on any branch because the current pipeline has a point of always making pushes to `gh-pages`. The reason why we need this is because `shippable` previously skipped builds for all branches without a `shippable.yml`; for CircleCI , we need to be a little more explicit. I don't think it will prevent any commits going in to `gh-pages`, it will just skip running the build, avoiding any unnecessary alerts and use of credits.

Not sure why this link is broken https://developers.gro-intelligence.com/CLEWS-38706-doc/api.html#groclient.GroClient.get_descendant? Is it related to this branch? The changes here shouldn't cause any breaks since the PR hasn't been merged, right?

PR #327, opened 2021-06-22
Add circleci status badge

kmetallic04:
Add a CircleCI status badge that indicates whether the current build is failing or passing, similar to what we have here: https://github.com/gro-intelligence/api-client/tree/CLEWS-37320-add-circleci-status-badge


MichaelWakahe (approved):

muzigao:
@kmetallic04  I think after i merged https://github.com/gro-intelligence/api-client/pull/324,  github automatically closed that branch so any further build would fail.

kmetallic04:
> @kmetallic04 I think after i merged #324, github automatically closed that branch so any further build would fail.

@muzigao I think there was a race condition where the automatic deletion happened in the period between which the build cloned `gh-pages` and the point when it attempted to push upstream. A simple retrigger did the trick. We should definitely try and minimize the length of this window (by e.g adding a git pull just before the auto-commit step).

PR #328, opened 2021-06-23
CLEWS-39481 Introduce get_ancestors function

franklinsagini:
Similar to `get_descendants`, introduce `get_ancestors` method that returns all ancestors for a given entity id.  

Usage

```
>>> from groclient import GroClient
>>> client = GroClient(API_HOST, ACCESS_TOKEN)

>>> client.get_ancestors('regions', 1107, 1, False)
[{'id': 0}, {'id': 11}, {'id': 100000004}, {'id': 100000008}, {'id': 100000009}, {'id': 100000010}, {'id': 100000012}, {'id': 100017849}]

>>> client.get_ancestors('regions', 13455, None, False)
[{'id': 0}, {'id': 11}, {'id': 1107}, {'id': 100000004}, {'id': 100000008}, {'id': 100000009}, {'id': 100000010}, {'id': 100000012}, {'id': 100000195}, {'id': 100017849}]

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -22,6 +22,8 @@ Basic Exploration
 
 .. automethod:: groclient.GroClient.search_for_entity
 
+.. automethod:: groclient.GroClient.get_ancestors
+
 .. automethod:: groclient.GroClient.get_descendant
 
 .. automethod:: groclient.GroClient.get_data_series
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -22,7 +22,7 @@ Basic Exploration
 
 .. automethod:: groclient.GroClient.search_for_entity
 
-.. automethod:: groclient.GroClient.get_ancestors
+.. automethod:: groclient.GroClient.get_ancestor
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: docs/api.rst [hunk 1]
```diff
@@ -22,6 +22,8 @@ Basic Exploration
 
 .. automethod:: groclient.GroClient.search_for_entity
 
+.. automethod:: groclient.GroClient.get_ancestors
```
muzigao: ahhh, just realized that we used singular name for `get_descendant` :( 
should we just call it `get_ancestor` ? any preference?  (sorry I know i put `ancestors` in original task  )

franklinsagini: I also thought about this, I thing `get_ancestor` is better, since it's consistent with the initial naming

muzigao (approved):
This looks good. exact same mirror functions as current `get_descendant()`.
I think we can take a separate clean up to consolidate region level and those separate `get_descendant_regions()` or `get_provinces()` functions

eliaschneider (commented):
LGTM

PR #330, opened 2021-06-28
CLEWS-38303 support show_history in get data calls

cn1036:
- deprecated 'show_available_date' (if a user includes such param in the query, we will not throw an error but the result remains the same regardless of the value of 'show_available_date')
- added an alias 'reporting_history' for 'show_revisions'
- introduced a new param 'complete_history'

**Status: ready for review**


groclient/client.py
```diff
@@ -963,11 +963,12 @@ def get_df(
...
-                False by default, meaning only the latest value for each period. If true, will return
-                all values for a given period, differentiated by the `reporting_date` field.
-            show_available_date : boolean, optional
-                False by default. If true, will return the available date of each data point.
+            reporting_history : boolean, optional
+                False by default, meaning only the latest value for each period. If true, will return all values for a given period 
+                which have a valid reporting_date.
+            complete_history : boolean, optional
+                False by default, meaning only the latest value for each period. If true, will return all values for a given period, 
+                even if there is no valid reporting_date. Will return values which are differentiated only be available_date.
```
cn1036: @JGrinst 
In the last sentence `... which are differentiated only be available_date.`, I was wonder if it should be 'differentiated only by'? And it's also possible for multiple data points to share the same available date. e.g. when their reporting dates are different.

api/client/samples/analysis_kits/stocks_to_use_price_model/stocks_to_use_price_model.ipynb
```diff
@@ -102,7 +102,7 @@
     "                                   'source_id': CME,\n",
     "                                   'frequency_id': 15,\n",
     "                                   'start_date': '2006-01-01',\n",
-    "                                   'show_revisions': True})\n",
+    "                                   'reporting_history': True})\n",
```
muzigao, line 105: should we just call "reporting_history" in API as well? ( last time I renamed it to show_reporting_date)

groclient/client.py
```diff
@@ -1012,11 +1012,12 @@ def get_df(
 
         Parameters
         ----------
-            show_revisions : boolean, optional
-                False by default, meaning only the latest value for each period. If true, will return
-                all values for a given period, differentiated by the `reporting_date` field.
-            show_available_date : boolean, optional
-                False by default. If true, will return the available date of each data point.
+            reporting_history : boolean, optional
```
muzigao, line 1015: (same for `get_data_points` below)
I think here it's better to elaborate that this parameter controls the source reporting history 
Also, I found here "value " is a bit confusing

maybe something like this? 
```
reporting_history: boolean, optional
False by default. If true, will return all reporting history from the source. 
```

groclient/client.py
```diff
@@ -1012,11 +1012,12 @@ def get_df(
...
+                False by default, meaning only the latest value for each period. If true, will return all values for a given period 
+                which have a valid reporting_date.
+            complete_history : boolean, optional
+                False by default, meaning only the latest value for each period. If true, will return all values for a given period,
```
muzigao: same here, I think `meaning only the latest value for each period` is a bit confusing , maybe we highlight `if true` part, explaining we're returning changing history from both source and Gro

groclient/client.py
```diff
@@ -1012,11 +1012,12 @@ def get_df(
...
         Parameters
         ----------
-            show_revisions : boolean, optional
-                False by default, meaning only the latest value for each period. If true, will return
-                all values for a given period, differentiated by the `reporting_date` field.
-            show_available_date : boolean, optional
-                False by default. If true, will return the available date of each data point.
+            reporting_history : boolean, optional
+                False by default, meaning only the latest value for each period. If true, will return all values for a given period 
+                which have a valid reporting_date.
```
muzigao: also I think we might be better keep `show_revisions` , but explaining it's deprecating and renamed to `reporting_history`

muzigao (approved):
Code change lgtm. Left comments for document

api/client/samples/analysis_kits/stocks_to_use_price_model/stocks_to_use_price_model.ipynb
cn1036: discussed offline. it requires a mandatory upgrade in the client side if we want to deprecate the old param in API endpoint
leaving it as a future TODO

groclient/client.py
JGrinst: Good point. So perhaps, instead:

reporting_history: boolean, optional
False by default. If true, will return all reporting history from the source.

complete_history: boolean, optional
False by default. If true, will return complete history of data points for the selection. This will include the reporting history from the source and revisions Gro has captured that may not have been released with an official reporting_date.

groclient/client.py
cn1036: added

groclient/client.py
cn1036: done

JGrinst (approved):

PR #332, opened 2021-07-01
document new param :available_since

franklinsagini:
Document newly supported param `available_since` by adding it as inline comments and updating querying-data docs


muzigao (approved):
LGTM

PR #335, opened 2021-07-21
CLEWS-40156 consolidating get_descendant_regions() into get_descendant(), updating get_ancestor() for regions

yohn-dezmon:
* consolidating get_descendant_regions() into get_descendant()
* updating get_ancestor() to include regional parameters and logic 
* updated unit tests with new parameters for each method

It would be good to test this in the staging environment to ensure that the region_level parameter and distance parameter are working correctly.


groclient/lib.py
```diff
@@ -606,70 +606,63 @@ def get_geojson(access_token, api_host, region_id, zoom_level):
...
 
 def get_ancestor(access_token, api_host, entity_type, entity_id, distance=None,
-                   include_details=True):
+                 include_details=True, ancestor_level=None, include_historical=True,):
     url = '/'.join(['https:', '', api_host, 'v2/{}/belongs-to'.format(entity_type)])
     headers = {'authorization': 'Bearer ' + access_token}
     params = {'ids': [entity_id]}
     if distance:
         params['distance'] = distance
+    elif ancestor_level:
```
franklinsagini: ancestor_level only applies to regions, so we should add another condition to check for entity type, otherwise the request will fail

groclient/lib.py
```diff
@@ -606,70 +606,63 @@ def get_geojson(access_token, api_host, region_id, zoom_level):
...
-    return [{'id': descendant_entity_id} for descendant_entity_id in descendant_entity_ids]
-
-
-def get_descendant_regions(access_token, api_host, region_id,
-                           descendant_level=None, include_historical=True, include_details=True):
-    url = '/'.join(['https:', '', api_host, 'v2/regions/contains'])
-    headers = {'authorization': 'Bearer ' + access_token}
-    params = {'ids': [region_id]}
-    if descendant_level:
+    elif descendant_level:
```
franklinsagini: same here, check for entity regions as well

groclient/client.py
```diff
@@ -766,6 +768,12 @@ def get_ancestor(
             True by default. Will perform a lookup() on each ancestor to find name,
             definition, etc. If this option is set to False, only ids of ancestor
             entities will be returned, which makes execution significantly faster.
+        ancestor_level : integer, optional
```
franklinsagini, line 777: add a comment in both methods stating when both distance and ancestor_level are specified, distance takes precedence

groclient/lib.py
yohn-dezmon: Ok definitely, will do, thank you!

groclient/lib.py
yohn-dezmon: right, thanks!

groclient/client.py
yohn-dezmon: Awesome will do, I wasn't sure how those two parameters would interact, so thanks for clearing that up.

groclient/client.py
```diff
@@ -750,21 +750,37 @@ def get_ancestor(
...
+        ancestor_level=None,
+        include_historical=True,
     ):
-        """Given an item, metric or region, returns all its ancestors i.e.
-        entities that "contain" in the given entity
+        """Given an item, metric, or region, returns all its ancestors i.e.
+        entities that "contain" in the given entity.
+
+        The `distance` parameter controls how many levels of ancestor entities you want to be
+        returned. Additionally, if you are getting the descendants of a given region, you can
```
muzigao: typo - should be `if you are getting the ancestors`

groclient/client.py
```diff
@@ -854,9 +886,13 @@ def get_descendant_regions(
         descendant_level=None,
         include_historical=True,
         include_details=True,
+        distance=None,
     ):
         """Look up details of all regions of the given level contained by a region.
 
+        This method is deprecated, and should be replaced by :meth:`~.get_descendant` which has
+        been updated to include the `descendant_level` and `include_historical` parameters.
```
franklinsagini, line 894: update get_provinces to use get_descendant

franklinsagini (approved):

groclient/client.py
yohn-dezmon: thanks! Updated

groclient/client.py
yohn-dezmon: Got it, thank you!

muzigao (approved):
LGTM!

PR #339, opened 2021-08-30
CLEWS-38710 check if required params are defined in get data call

cn1036:


eliaschneider (approved):
LGTM

groclient/lib.py
```diff
@@ -542,9 +542,16 @@ def list_of_series_to_single_series(series_list, add_belongs_to=False, include_h
 
 
 def get_data_points(access_token, api_host, **selection):
+    logger = get_default_logger()
```
muzigao, line 545: there's also `get_data_points_generator()` in client.py (it's a bit legacy but we have to maintain both places right now

muzigao (approved):

PR #340, opened 2021-11-05
DELI-132: remove old version of similar region sample

muzigao:
Removing old similar_region code from github as we're launching the new version


cn1036 (approved):

tolothy:
:wave:

PR #341, opened 2022-02-15
DELI-276 include historical refactor for get_descendant()

yohn-dezmon:
Previously get_descendant() made a call to `v2/regions/contains` to get the descendant ids for a given region, then if include_historical was set to false, it would do batch lookups of the descendant ids against the `v2/regions` endpoint. After getting a response, it would filter the data for the regions that had include_historical set to false.

This PR is directly related to the DELI-276 PR on the `gro/api` code base, where the includeHistorical parameter was added to the `/v2/reginos/contains` endpoint as an optional parameter. Thus, now the filtering for the historical regions is done at the DB level, and unnecessary processing at the python API client is no longer necessary.

### DIFF
--- a/groclient/lib.py
+++ b/groclient/lib.py
@@ -676,7 +676,6 @@ def get_descendant(access_token, api_host, entity_type, entity_id, distance=None
             params['distance'] = -1
 
     params['includeHistorical'] = include_historical
-
     resp = get_data(url, headers, params)
     descendant_entity_ids = resp.json()['data'][str(entity_id)]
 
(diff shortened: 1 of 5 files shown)


groclient/lib.py
```diff
@@ -675,20 +675,16 @@ def get_descendant(access_token, api_host, entity_type, entity_id, distance=None
         else:
             params['distance'] = -1
 
+    if not include_historical:
```
cn1036: minor: can simply use `params['includeHistorical'] = include_historical`

cn1036 (changes requested):
one minor comment, otherwise lgtm

groclient/lib.py
yohn-dezmon: good call, updating now.

cn1036 (approved):

cn1036:
waiting for the prod release

muzigao-gro (approved):

PR #342, opened 2022-03-03
copied changes from DELI-276-include-historical-refactor

yohn-dezmon:
Previously get_descendant() made a call to v2/regions/contains to get the descendant ids for a given region, then if include_historical was set to false, it would do batch lookups of the descendant ids against the v2/regions endpoint. After getting a response, it would filter the data for the regions that had include_historical set to false.

This PR is directly related to the DELI-276 PR on the gro/api code base, where the includeHistorical parameter was added to the /v2/regions/contains endpoint as an optional parameter. Thus, now the filtering for the historical regions is done at the DB level, and unnecessary processing at the python API client is no longer necessary.


jsbrucker:
Duplicate

PR #347, opened 2022-03-15
[INFRA-575] create workflows for merges & releases

eric-gro:
Create separate workflows for:
- pull requests
- merges into development branch
- release tags


.circleci/config.yml
```diff
@@ -77,42 +78,105 @@ jobs:
...
+          <<: *build-docs
+      - run:
+          <<: *test-docs
+      - run:
+          <<: *publish-docs
+      - run:
+          <<: *build-package
       # Publish new releases to PyPI.
       # We only check whether this is a tagged build and if so, publish to pypi
       # Check that $CIRCLE_TAG is not a null string
```
jsbrucker: Does this comment need to be updated?

.circleci/config.yml
```diff
@@ -77,42 +78,105 @@ jobs:
...
+      - run:
+          <<: *test-docs
+      - run:
+          <<: *publish-docs
+      - run:
+          <<: *build-package
       # Publish new development updates to TestPyPI whenever changes are merged
       # to development.
       # We only want to publish a new package when PRs are actually accepted and
       # merged, which is why we check $CIRCLE_PULL_REQUEST.
```
jsbrucker: Ditto

.circleci/config.yml
```diff
@@ -77,42 +78,105 @@ jobs:
...
+          filters:
+            branches:
+              only: "/^development$/"
+
+  release:
+    jobs:
+      - release-build:
+          filters:
+            branches:
+              ignore: "/.*/"
```
jsbrucker, line 179: Does this signify ignore changes to every branch?
Combined with the tags line does it mean only run when a tag is added?

.circleci/config.yml
eric-gro: Yes.. needs an update. Need CircleCI to build for a PR first though. Which is has.

.circleci/config.yml
eric-gro: Yes, combined it means only run with every new tag.

eric-gro:
@muzigao do we want to push docs every time we merge to development or every time we create a new release tag?

gro-raywan (approved):
LGTM

John-Desmond (approved):
LGTM

muzigao-gro (approved):

muzigao-gro:
> @muzigao do we want to push docs every time we merge to development or every time we create a new release tag?

Right now the doc site is by default showing development version. There's an selection you can toggle between release tags I think.  

We might want to revisit the behavior of dev doc (by default showing latest Tag), but can we keep the current behavior until we have the bandwidth?

fantastic-mr-fox (approved):
LGTM

Mbui-John (approved):

PR #348, opened 2022-04-04
added tilde to Sao Paulo; updated time of year (late 2021)

nataliegable-gro:
Very small change so that the notebook runs as expected.

Changes:
* Added tilde so it reads São Paulo instead of Sao Paulo (was getting the wrong data series without)
* Specified an updated time of the year, to use the features that are used in the last period of the year (data from late 2021) instead of the middle.


elia-gro:
LGTM

elia-gro (approved):

yi-zhao-gro-intelligence (approved):
Change looks good! 
It's difficult to prevent this type of error in the future, unless we hard code the region_id instead, but that also beats the purpose of making it a demonstration on how to use the API client. So let's keep it the way it is until further product decision on whether to include notebooks like this one on the public repo.

PR #350, opened 2022-06-02
Add crop weighting api

jpaye:
The API team has deployed the new crop weighting API, and I think it would be useful to include it in the Gro API client. This standardizes how the endpoint is called and avoids having to directly use the requests library.

Notes:

- I think this change does not affect permissioning, since calling the area weighting endpoint will still be admin-only (and thus internal-only) until we decide otherwise
- The broadest change below is changing `requests.get` to the more general `requests.request` in the `get_data` function. Did this so that we can send POST requests with the `get_data` logic, not just GET requests
- I have tried to follow the existing structure as much as possible for testing--am the first to admit that the tests as written below don't test all that much, so if anyone has ideas for more thorough tests please let me know


John-Desmond:
I think it's unrelated to your changes, but looks like we have some broken links in the docs of this repo that are causing the circleci build to fail: https://app.circleci.com/pipelines/github/gro-intelligence/api-client/272/workflows/0aedc091-22ab-41ca-aada-abfa45a8a6ed/jobs/293

John-Desmond:
^ This happened to me once before, if I can I'll try to find the PR where I updated one of the broken links

groclient/lib.py
```diff
@@ -387,6 +387,13 @@ def get_data_series(access_token, api_host, **selection):
     except KeyError:
         raise Exception(resp.text)
 
+def weight_area(access_token, api_host, series_name, weights):
+    url = '/'.join(['https:', '', api_host, 'area-weighting'])
+    headers = {'authorization': 'Bearer ' + access_token}
+    params = {'seriesName' : series_name}
+    resp = get_data(url, headers, params = params, request_method = "post", request_data = {'weights' : json.dumps(weights)})
```
John-Desmond: nit: PEP8 says there shouldn't be spaces between parameters and keywords, e.g. should be `params=params`

John-Desmond: I'm interested if you tested this against the actual Gro API endpoint and if it worked. 
```
{'weights' : json.dumps(weights)}
```
I _think_ it should be `{"weights": {"1215": 0.15}}` , not `{'weights': "{'1215': 0.15}"}`

John-Desmond: this is from the climate navigator code: `areaSeries = {'weights': areaSeries.to_json()}`, doing some local testing now to see if `to_json()` outputs a dict or a str. My guess would be a dict.

John-Desmond: ok looks like I was wrong, from the to_json() docs: `Convert the object to a JSON string.` I need to do some thinking to see if my most recent changes will work with this format. I tried to change the format on the Gro API side to more closely match the input into the fastapi endpoints. Looking into it now and will post back here once I've verified.

John-Desmond: @jpaye I reverted the changes, the Gro API will now expect the format:
```
{'weights':"{\"14114\":0.15}"}
```
as is used here and in the Climate Risk Navigator.

John-Desmond (approved):

groclient/lib.py
jpaye: ty, fixed

danny-droll (approved):

groclient/client.py
```diff
@@ -1604,6 +1604,43 @@ def search_for_entity(self, entity_type, keywords):
             )
             return result["id"]
 
+    def weight_area(self, series_name, weights):
```
bernardigro: i think sticking to the "get_x" naming convention would make it more obvious to what the function actually does and fit with the existing functions like get_data_points, get_data_series, etc. "get_area_weighted_series"?

jpaye: Agree, updated

bernardigro (approved):

groclient/lib.py
```diff
@@ -160,7 +160,7 @@ def convert_value(value, from_convert_factor, to_convert_factor):
     ) / to_convert_factor.get("factor")
 
 
-def get_data(url, headers, params=None, logger=None, stream=False):
+def get_data(url, headers, params=None, logger=None, stream=False, request_method="get", request_data=None):
     """General 'make api request' function.
 
     Assigns headers and builds in retries and logging.
```
muzigao-gro, line 166: can we add those two new arguments in comments?

jpaye: Yes, added

PR #351, opened 2022-06-09
fix data download bug

nataliegable-gro:
Using the old notebook, it was specifying the region ID to be the district of Sao Paulo (versus the province, which is what we want). I changed the region ID and also changed the notebook to utilize `GroClient` instead of `CropModel`, which allows for the use of functions like `get_ancestor`.


nataliegable-gro:
<p>Added comment</p>

---

 <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/351/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

nataliegable-gro:
<p>The way we were downloading data before (adding series to the client, getting the dataframe of all the series, then filtering) was resulting in us having both GPM and GFS in the same dataframe (both with the same item/metric for precip). Because of this, the left merge worked because we already had GPM data in the <code>raw_gfs1</code> dataframe. Now, since we are only getting the GFS data in the <code>raw_gfs1</code> dataframe, we need to make this a right merge. This results in a dataframe where we have GPM data for all days available, and then we're padding the NaN days with GFS data as desired  </p>

---

 <a href='https://app.reviewnb.com/gro-intelligence/api-client/pull/351/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

yi-zhao-gro-intelligence (approved):

muzigao-gro (approved):

PR #353, opened 2022-08-03
GAIA-9469 Proxy Enviornment Variables for Tornado package

John-Desmond:
When an end user is using the api-client and they have to access the internet via a proxy, currently the Tornado `AsyncHTTPClient` was throwing errors because it requires `proxy_host` and `proxy_port` as parameters when being accessed via a proxy. 

This PR Updates `GroClient` to provide optional parameters, `proxy_host` and `proxy_port`, which are then passed to the `AsyncHTTPClient` which also is configured as a `CurlAsyncHTTPClient` class (otherwise the proxy variables won't work).

I tested this locally by installing `mitmproxy` with homebrew, and updating my wifi settings such that all requests/responses are routed through it. It runs on `localhost:8080`. I tested that the code works with and without setting the `PROXY_HOST` and `PROXY_PORT` environment variables.

Confirmed that setting the username and password was succesful with `mitmproxy --proxyauth any`:
Screen Shot 2022-08-03 at 1 04 43 PM

Also this required installing `pycurl` so I'll need to update requirements.txt

```
$ pip install pycurl 
$ brew info openssl 
< set environment variables that it specifies >

For compilers to find openssl@3 you may need to set:
  export LDFLAGS="-L/usr/local/opt/openssl@3/lib"
  export CPPFLAGS="-I/usr/local/opt/openssl@3/include"

For pkg-config to find openssl@3 you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/openssl@3/lib/pkgconfig"
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ chardet
 future
 numpy
 pandas
+pycurl
 python-dateutil
 pytz
 requests
--- a/api/client/samples/batch_queries/batch_queries.py
+++ b/api/client/samples/batch_queries/batch_queries.py
@@ -1,8 +1,5 @@
 import os
-# change this import so it doesn't import from groclient but rather from the relative directory
-# from groclient import GroClient
-# you need to add /api-client to the PYTHONPATH envvar
-from groclient.client import GroClient
+from groclient import GroClient
 
 def main():
     # set up the Batch Client, same as normal Client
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -4,6 +4,7 @@ chardet
 future
 numpy
 pandas
+pycurl
```
muzigao-gro, line 7: do we need this?

loganmcdougallgro (approved):

requirements.txt
John-Desmond: Yeah, it became a requirement once I used this line (which was also required for the new proxy variables): 
```
AsyncHTTPClient.configure("tornado.curl_httpclient.CurlAsyncHTTPClient")
```

John-Desmond: otherwise you'll get the error: 
```
ModuleNotFoundError: No module named 'pycurl'
```

muzigao-gro (approved):
code change LGTM!

PR #362, opened 2022-10-18
Change the log level from error to warning.

prabesh-paudel:
Changed the log level to warning where there is a bad request.
I scanned the client and could find these two places where we could change the log level from error to warning.
I also scanned if there are any places where we should change from warning level to error level log but none such case exist.

### DIFF
--- a/groclient/client.py
+++ b/groclient/client.py
@@ -347,7 +347,7 @@ class GroClient(object):
                 if len(missing_params) == 1
                 else ', '.join(missing_params[:-1]) + ' and ' + missing_params[-1] + ' are'
             )
-            self._logger.error(message)
+            self._logger.warning(message)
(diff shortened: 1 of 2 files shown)


groclient/client.py
```diff
@@ -347,7 +347,7 @@ def get_data_points_generator(self, **selection):
                 if len(missing_params) == 1
                 else ', '.join(missing_params[:-1]) + ' and ' + missing_params[-1] + ' are'
             )
-            self._logger.error(message)
```
John-Desmond, line 350: I'm a bit confused as to why we need logging sent to Kibana from the `api-client` at all, but LGTM.

John-Desmond (approved):

cn1036 (approved):

PR #363, opened 2022-10-24
GAIA-11568 add param coverage_thresold to data calls

cn1036:
The API change is addressed in https://github.com/gro-intelligence/gro/pull/3367.
Note: The filtering by coverage feature is for internal use only. the documentation is not updated so that external users should not be aware of this change.

### DIFF
--- a/groclient/lib.py
+++ b/groclient/lib.py
@@ -372,7 +372,7 @@ def get_data_call_params(**selection):
             params['showHistory'] = value
         elif key in ('show_revisions', 'reporting_history'):
             params['showReportingDate'] = value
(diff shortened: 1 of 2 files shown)


John-Desmond (approved):
LGTM, but should we add a unit test for this new use case? Or is a unit test here unnecessary since there is no change in logic on the api-client side?

cn1036:
> LGTM, but should we add a unit test for this new use case? Or is a unit test here unnecessary since there is no change in logic on the api-client side?

I didn't include any

cn1036:
> LGTM, but should we add a unit test for this new use case? Or is a unit test here unnecessary since there is no change in logic on the api-client side?

unit test added

PR #365, opened 2022-12-20
Add show_metadata to get_data_points() documentation

John-Desmond:
@NickHilton pointed out on the #internal_api_support channel that this optional parameter is not specified in the `get_data_points()` documentation. This PR adds the `show_metadata` parameter to the get_data_points() method's documentation. 

**Tested**

Verified the parameter works with the given endpoint: 
```
data_points = client.get_data_points(**{'metric_id': 170037,
                          'item_id': 274,
                          'region_id': 1215,
                          'frequency_id': 9,
                          'source_id': 32,
                          'start_date': '2001-01-01',
                          'unit_id': 284,
                          'show_metadata': True})
print(data_points[0])
```
Relevant output:
```
{ ...
'metadata': {'comment': 'this is a simulation produced on 2019-10-31', 'ecdf_absolute': 0.7605085419149781, 'conf_interval': 2.759466933710244}, 'input_unit_id': 284,
...
}
```


groclient/client.py
```diff
@@ -1300,6 +1300,8 @@ def get_data_points(self, **selections):
             True by default, will include historical regions that are part of your selections
         available_since : string, optional
             Fetch points since last data retrieval where available date is equal to or after this date
+        show_metadata : bool, optional
```
muzigao-gro, line 1305: @cn1036 do you remember why we didn't show this flag before?

cn1036: I can't recall. this param exists in lib.py since 2020. in my opinion, there is no harm to expose it to the end users.

cn1036 (approved):

NickHilton (approved):
Nice!

PR #369, opened 2023-01-25
GAIA-16899 Multiple region ids for `get_area_weighted_series()`

John-Desmond:
Solution:
* update region_id to accept either an int or list of ints

Tested:
* updated unit tests for `lib` and `client`

@jpaye used the following steps to do an integration test:
```
python3 -m poetry build
python3 -m pip uninstall groclient
python3 -m pip install groclient-0.0.0-py3-none-any.whl
client.get_area_weighted_series("GDI_daily", ["Corn"], [13066, 13065])
```

### DIFF
--- a/groclient/client.py
+++ b/groclient/client.py
@@ -1703,7 +1703,7 @@ class GroClient(object):
         """
         return lib.get_area_weighting_weight_names(self.access_token, self.api_host)
 
-    def get_area_weighted_series(self, series_name, weight_names, region_id=None, region_ids=[], method='sum', latest_date_only=False):
+    def get_area_weighted_series(self, series_name, weight_names, region_id, method='sum', latest_date_only=False):
         """Compute weighted average on selected series with the given weights.
 
         Returns a dictionary mapping dates to weighted values.
--- a/groclient/client.py
+++ b/groclient/client.py
@@ -1703,7 +1703,7 @@ class GroClient(object):
         """
         return lib.get_area_weighting_weight_names(self.access_token, self.api_host)
 
-    def get_area_weighted_series(self, series_name, weight_names, region_id, method='sum', latest_date_only=False):
+    def get_area_weighted_series(self, series_name, weight_names, region_id=None, region_ids=[], method='sum', latest_date_only=False):
         """Compute weighted average on selected series with the given weights.
 
         Returns a dictionary mapping dates to weighted values.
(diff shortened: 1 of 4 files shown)


jpaye (approved):
tested locally, looks good!

groclient/client.py
```diff
@@ -1716,8 +1716,8 @@ def get_area_weighted_series(self, series_name, weight_names, region_id, method=
         weight_names: list of strs
             List of weight names that will be used to weight the provided series. e.g. ['Barley (ha)', 'Corn (ha)']
             For getting the full list of valid weight names, please call :meth:`~.get_area_weighting_weight_names`
-        region_id: integer
-            The region for which the weighted series will be computed
+        region_id: integer or list of integers
```
HonghuiLu2021, line 1722: Rename to region_id`s`? Are there any restrictions, like equal to or above level5?

HonghuiLu2021: Also, would be nice to add type annotations.

HonghuiLu2021 (commented):
Does this change break existing clients? The interface does not seem to be backward compatible.

John-Desmond:
> Does this change break existing clients? The interface does not seem to be backward compatible.

@HonghuiLu2021 yes this is backward compatible. Otherwise the unit tests would fail. In addition @jpaye performed integration tests by building the PyPI image locally.

groclient/client.py
John-Desmond: yeah I was considering renaming the variable, but I thought if any end user was specifying the parameter like `region_id=1215` then this would be a breaking change. That use case is probably not very likely though (e.g. most users are probably just using positional arguments and not named arguments).

John-Desmond: as far as restrictions go I'm not entirely sure, I think if you request a region below L5 the endpoint will return an error response, but that's not new behavior

John-Desmond: yeah so for example, I took Minneapolis (id: 1001236, level=6) and ran it against the existing endpoint and this returns a 500 response: 
```
area-weighting?weightNames=Wheat&seriesName=NDVI_8day&regionId=1001236
...
{
    "statusCode": 500,
    "error": "Internal Server Error",
    "message": "An internal server error occurred"
}
```
I can update the comment to make it clear that regions below level 5 are not supported

John-Desmond: updated with type hints and a comment specifying supported region levels

jpaye: I think we should avoid breaking changes here-- if we rename `region_id` to `region_ids` we'll be breaking any code that currently uses the parameter. 

Agree that it's an inconsistency, but it has precedent in `get_data_points` ([link](https://developers.gro-intelligence.com/api.html#groclient.GroClient.get_data_points))

HonghuiLu2021 (approved):
LGTM. Thanks!

groclient/client.py
HonghuiLu2021: Ah, I didn't realized region_id is a union. Thanks for the update.

PR #374, opened 2023-04-03
GAIA-20153- Add support for v2-prime api call

prabesh-paudel:
**Goal/Problem**
Add `v2prime` API call support to python SDK.

**Solution**
Create a new experimental class that has function to call `v2Prime` endpoint.

**Tested**
Tested in local, able to get same data from API (from postman) and SDK.
```
exp = Experimental('localhost:3000', token)
exp_output = exp.get_data_points(**{'metric_id': 2540047,
                                    'item_ids': 3457,
                                    'region_ids': [100023971, 100023990],
                                    'frequency_id': 1,
                                    'source_id': 26,
                                    'start_date': '2021-12-20',
                                    'end_date': '2021-12-22',
                                    'stream': True,
                                    })
```

### DIFF
--- a/groclient/experimental.py
+++ b/groclient/experimental.py
@@ -33,7 +33,6 @@ class Experimental(GroClient):
                                     'source_id': 26,
                                     'start_date': '2021-12-20',
                                     'end_date': '2021-12-21',
-                                    'coverage_threshold': 0.9,
                                     'stream': True
                                     })
             Returns:
@@ -79,7 +78,7 @@ class Experimental(GroClient):
                 "meta": {
                     "version": "local",
                     "copyright": "Copyright (c) Gro Intelligence",
-                    "timestamp": "Mon, 03 Apr 2023 18:22:10 GMT"
+                    "timestamp": "Mon, 03 Apr 2023 18:32:53 GMT"
                 }
             }
         Parameters
@@ -103,6 +102,7 @@ class Experimental(GroClient):
         end_date : string, optional
             All points with start dates equal to or before this date
         coverage_threshold: float, optional
+            custom threshold on the coverage of geospatial data
         stream: bool, optional
 
         Returns
--- a/groclient/experimental.py
+++ b/groclient/experimental.py
@@ -1,0 +1,115 @@
+from groclient.client import GroClient
+from groclient import lib
+
+
+class Experimental(GroClient):
+    """
+    Experimental class to consume v2prime data.
+    """
+    # Query parameters other than these are ignored.
+    ALLOWED_SELECTIONS = {
+        'item_ids',
+        'region_ids',
+        'metric_id',
+        'partner_region_ids',
+        'frequency_id',
+        'source_id',
+        'start_date',
+        'end_date',
+        'unit_Id',
+        'stream',
+        'coverage_threshold',
+                          }
+
+    def get_data_points(self, **selections):
+        """
+        Get all the data points for a given selection from v2prime/data.
+        Similar to groclient's get_data_points() for v2/data.
+        Example:
+            experimental_client.get_data_points(**{'metric_id': 2540047,
+                                    'item_ids': [3457],
+                                    'region_ids': [100023971, 100023990],
+                                    'frequency_id': 1,
+                                    'source_id': 26,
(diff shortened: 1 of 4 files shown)


groclient/experimental.py
```diff
@@ -0,0 +1,115 @@
...
+            and the "partner_region" would be the importer. For most series, this can be excluded
+            or set to 0 ("World") by default.
+        source_id : integer
+        frequency_id : integer
+        unit_id : integer, optional
+        start_date : string, optional
+            All points with end dates equal to or after this date
+        end_date : string, optional
+            All points with start dates equal to or before this date
+        coverage_threshold: float, optional
```
prabesh-paudel, line 89: Is the `coverage_threshold` supposed to be between 0 and 1? @cn1036

groclient/lib.py
```diff
@@ -576,6 +576,39 @@ def get_data_points(access_token, api_host, **selection):
...
+        if key in allowed_selections:
+            params[groclient.utils.str_snake_to_camel(key)] = value
+    return params
+
+
+def get_data_points_v2_prime(access_token, api_host, allowed_selections, **selection):
+    logger = get_default_logger()
+    headers = {'authorization': 'Bearer ' + access_token}
+    url = '/'.join(['https:', '', api_host, 'v2prime/data'])
+    params = parse_params(allowed_selections, **selection)
```
cn1036: checking for required params can be combined with parse_params()
if we need to hardcode the query params in SDK anyway then why keeping 2 copies in both constant.py and experimental.py? we could have simply have one copy in constant and mark the required fields

prabesh-paudel: yeah I should do that. I wanted to follow make it similar to `get_data_points` for v2data but it seems thats not the best way.
Good catch, will do.

groclient/experimental.py
```diff
@@ -0,0 +1,115 @@
+from groclient.client import GroClient
+from groclient import lib
+
+
+class Experimental(GroClient):
+    """
+    Experimental class to consume v2prime data.
+    """
+    # Query parameters other than these are ignored.
+    ALLOWED_SELECTIONS = {
```
muzigao-gro: reading lib.get_data_points() again, I think we don't need to declare all allowed selections here , we should make it flexible for python client. only 5 tuples are required, the others are optional

groclient/experimental.py
```diff
@@ -0,0 +1,115 @@
...
+            custom threshold on the coverage of geospatial data
+        stream: bool, optional
+
+        Returns
+        -------
+
+        dict of data_series containing data points and its description as shown in above example.
+
+        """
+        return lib.get_data_points_v2_prime(self.access_token, self.api_host, Experimental.ALLOWED_SELECTIONS,
```
muzigao-gro: and I think it's better to not passing a constant to lib.get_data_points_v2_prime, we can check user selections within lib.get_data_points_v2_prime function

groclient/lib.py
```diff
@@ -576,6 +576,39 @@ def get_data_points(access_token, api_host, **selection):
...
 
+def parse_params(allowed_selections, **selections):
+    params = {}
+    for key, value in list(selections.items()):
+        if key in allowed_selections:
+            params[groclient.utils.str_snake_to_camel(key)] = value
+    return params
+
+
+def get_data_points_v2_prime(access_token, api_host, allowed_selections, **selection):
```
muzigao-gro: again, we should not be passing allowed_selections

groclient/experimental.py
muzigao-gro: another point is - should let API to manage query parameter validation so we don't have to replicate that in python sdk

muzigao-gro (approved):

muzigao-gro:
code change LGTM, pending answers regarding coverage

groclient/experimental.py
cn1036: yes

groclient/lib.py
prabesh-paudel: done

groclient/experimental.py
prabesh-paudel: done

cn1036 (approved):

PR #377, opened 2023-04-07
Gaia 20182 Increase read chunksize on iterlines() and remove chunksize from  stream_data_series()

prabesh-paudel:
Goal:
The sdk was slow and had adjustable `chunk_size` (not the read chunk size, but the param to determine how many dataseries to be yielded on each iteration of `stream_data_series()`) parameter.

Solution:
Increase the read chunksize of ` iter_lines()` to 4 MB. And remove the `chunk_size` param from `stream_data_series()` while setting it fixed at `10k` dataseries.

Tested:
Tested in local.


groclient/client.py
```diff
@@ -654,7 +654,7 @@ def get_data_series(self, **selection):
         """
         return lib.get_data_series(self.access_token, self.api_host, **selection)
 
-    def stream_data_series(self, chunk_size=10000, **selection):
+    def stream_data_series(self, **selection):
```
HonghuiLu2021, line 657: Remove comments at lines 663, 664.

HonghuiLu2021 (approved):
LGTM. Need to update some comments.

groclient/client.py
prabesh-paudel: Done.

PR #378, opened 2023-04-19
Gaia 20447 - Add developers docs for experimental v2 prime get_data_points function

prabesh-paudel:
**Goal**
We have `Experimental` class that we can use to access the v2prime data.
We want to add the documentation to the function `get_data_points` of the `Experimental` class.

**Solution**
Add the documentation and make it visible under [https://developers.gro-intelligence.com/api.html](https://developers.gro-intelligence.com/api.html).

**Tested**
Works on my local. (Updated following image)
image

Notes to reviewer:

Seems like if you tag a commit, the `CD` automatically releases a new version? I was tagging a commit in my branch to see if only the docs gets published but ended up having a code release. Has no code change in that commit so it isn't a problem. Hopefully the prod release includes the permission for v2 prime tomorrow. Will merge this and release a new version manually after review.

### DIFF
--- a/groclient/experimental.py
+++ b/groclient/experimental.py
@@ -7,6 +7,7 @@ class Experimental(GroClient):
 
     def get_data_points(self, **selections):
         """Get all the data points for a given selection from /v2prime/data.
+
         Note - This function is a part of an experimental class and subject to change.
 
         Parameters
@@ -40,18 +41,19 @@ class Experimental(GroClient):
 
             Example::
 
-            exp_client = Experimental(api.gro-intelligence.com, GROAPI_TOKEN)
+            exp_client = Experimental("api.gro-intelligence.com", GROAPI_TOKEN)
             exp_client.get_data_points(
-            **{
-            'metric_id': 2540047,
-            'item_ids': [3457],
-            'region_ids': [100023971, 100023990],
-            'frequency_id': 1,
-            'source_id': 26,
-            'start_date': '2021-12-20',
-            'end_date': '2021-12-21',
-            'stream': False
-            })
+                **{
+                    'metric_id': 2540047,
+                    'item_ids': [3457],
+                    'region_ids': [100023971, 100023990],
+                    'frequency_id': 1,
+                    'source_id': 26,
+                    'start_date': '2021-12-20',
+                    'end_date': '2021-12-21',
+                    'stream': False
+                    }
+            )
             Returns:
             {
                 "data_series": [
--- a/groclient/experimental.py
+++ b/groclient/experimental.py
@@ -10,6 +10,7 @@ class Experimental(GroClient):
     def get_data_points(self, **selections):
         """
         Get all the data points for a given selection from /v2prime/data.
+        Note: This function is a part of an experimental class and subject to change.
 
         Example:
             experimental_client.get_data_points(**{'metric_id': 2540047,
@@ -27,7 +28,6 @@ class Experimental(GroClient):
                     {
                         "data_points": [
                             {
-                                "timestamp_sec": "1640044800",
                                 "value": 33.20465087890625,
                                 "start_timestamp": "1639958400",
                                 "end_timestamp": "1640044800"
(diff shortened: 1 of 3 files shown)


John-Desmond:
with regards to the tagging of a commit leading to a release, can you open a new bug ticket for that so we can address it later?

John-Desmond (approved):

John-Desmond:
Also were you able to find a way to see the documentation locally?

prabesh-paudel:
> Also were you able to find a way to see the documentation locally?

yes. 
cd to  `api-client/docs`
`sphinx-build -b html .  build/`
would create and save documents to the `api-client/docs/build`.

groclient/experimental.py
```diff
@@ -3,31 +3,65 @@
...
+        source_id : integer
+        frequency_id : integer
+        unit_id : integer, optional
+        start_date : string, optional
+            All points with end dates equal to or after this date
+        end_date : string, optional
+            All points with start dates equal to or before this date
+        coverage_threshold: float, optional
+            Custom threshold on the coverage of geospatial data. Value should be between 0 and 1.
+        stream: bool, optional
```
muzigao-gro: let's remove stream for now - it's not really making any difference to end users , we can add it back when we support real streaming between API & client

groclient/experimental.py
```diff
@@ -3,31 +3,65 @@
...
+            exp_client.get_data_points(
+                                    **{
+                                        'metric_id': 2540047,
+                                        'item_ids': [3457],
+                                        'region_ids': [100023971, 100023990],
+                                        'frequency_id': 1,
+                                        'source_id': 26,
+                                        'start_date': '2021-12-20',
+                                        'end_date': '2021-12-21',
+                                        'stream': False
```
muzigao-gro: same here

groclient/experimental.py
```diff
@@ -3,31 +3,65 @@
...
+            partner_region refers to an interaction between two regions, like trade or
+            transportation. For example, for an Export metric, the "region" would be the exporter
+            and the "partner_region" would be the importer. For most series, this can be excluded
+            or set to 0 ("World") by default.
+        source_id : integer
+        frequency_id : integer
+        unit_id : integer, optional
+        start_date : string, optional
+            All points with end dates equal to or after this date
+        end_date : string, optional
```
muzigao-gro, line 39: looks like we should remove `equal` for both start & end date

see https://github.com/gro-intelligence/gro/blob/main/storage/data_access/src/core_dbs/queries/serving_v2_queries.rs#L347

groclient/experimental.py
```diff
@@ -3,31 +3,65 @@
...
 class Experimental(GroClient):
-    """
-    Experimental class to consume v2prime data.
-    """
+    """Experimental class to consume v2prime data."""
 
     def get_data_points(self, **selections):
-        """
-        Get all the data points for a given selection from /v2prime/data.
+        """Get all the data points for a given selection from /v2prime/data.
```
muzigao-gro: nit:  can we change introduction to:

```
The experimental client will introduce a range of experimental functions with better user experience. While you will be able to access better performance and new features at an early stage, you should be aware that things might change (e.g response format).

This function is a mirror of existing :meth:`~get_data_points`, but with limited scope, for example:
- ontology expansion is under development 
- "Gro derived on-the-fly" is under development
- Many sources are still under migration (please refer to internal confluence page for source migration timeline)
```

muzigao-gro (approved):
left some minor comments

groclient/experimental.py
prabesh-paudel: Thanks. Done.

groclient/experimental.py
prabesh-paudel: ✔

groclient/experimental.py
prabesh-paudel: Thank you. Added this.

groclient/experimental.py
prabesh-paudel: Great catch thanks.

prabesh-paudel:
Now the docs looks like this:
image

PR #381, opened 2023-07-07
Add new get_df function to api doc

muzigao-gro:
in previous PR which we added new experimental.get_df() function for v2prime endpoint
I forgot to add new function to the doc

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -87,3 +87,4 @@ Experimental
 =============
 
 .. automethod:: groclient.Experimental.get_data_points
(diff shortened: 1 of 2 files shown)


docs/api.rst
```diff
@@ -87,3 +87,4 @@ Experimental
 =============
 
 .. automethod:: groclient.Experimental.get_data_points
+.. automethod:: groclient.Experimental.get_df
```
prabesh-paudel, line 90: The `get_df` function is under `groclient.GroClient`. Or am I missing something here?

PR #384, opened 2023-10-03
GAIA-26821-remove option 1 and 2 to get access token from docs

prabesh-paudel:
Goal/Problem:
Since we are moving to auth0, we don't want users to hit the api to get the access tokens (we don't store access tokens on gro db) . 
This PR only deletes the information (on documentation) on how you can get access token from command line and gro sdk `get_access_token` function.

Solution:
Remove the option 2 and option 3 from docs.

Tested:
Locally tested -  the markup doesn't show option 2 and 3.

### DIFF
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -11,10 +11,8 @@ Retrieving a token
 ==================
 
 Note that your account needs to be activated for API access before you will be able to retrieve a token. See https://gro-intelligence.com/products/gro-api for more info regarding unlocking API access for your account.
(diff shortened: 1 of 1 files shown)


muzigao-gro (approved):

daire-gro:
Can you also update further down in the document where it says "enter a password" and just remove "password or" so it says "if you don't want to enter a token each time" etc.

daire-gro (approved):
Looks good, thanks!  Can you also update where it says "enter a password" further down the file? thanks!

prabesh-paudel:
@daire-gro thanks for pointing that out. Updated it.

PR #387, opened 2023-12-19
GAIA-29800-Remove the get_access_token from sugar.py

prabesh-paudel:
**Goal/Problem**
The `get_access_token` is deprecated and will be removed. 
Remove the reference of the `get_access_token` from `sugar.py` and adjust its code accordingly.
This seems to be a sample file.

Tested
Tested in local by having --token flag and token set in `GROAPI_TOKEN`, both are able to get back the data.

Notes to reviewer:


John-Desmond (approved):

cn1036 (approved):
approved to unblock the release
but I think it would be better to let the corresponding team(owner of the script) know about this change.

PR #391, opened 2024-03-27
update stocks to use notebook

bernardigro:
the stocks to use vs price notebook was failing due to a change in search results.


danny-droll (approved):
lgtm

cn1036 (approved):

